PR #2, opened 2024-05-14
New monorepo layout

TheByronHimes:


mephenor:
Tests are broken without real changes from before, not exactly sure why.
Would just merge current state and cover all missing workflows first, then fix service level code.
Readmes and service level pyproject will need some additional work still that I also want to push to the end of the current queue.

mephenor (approved):

PR #18, opened 2024-06-10
Outbox publisher (GSI-773)

TheByronHimes:
This PR replaces the normal event publisher with the newer outbox dao publisher. When a file ID is specified for deletion, the File Deletion core layer inserts the file ID into the database and immediately publishes the data as an outbox-style event (meaning the extra `__metadata__` stored in the mongodb document is added to the kafka event). 

This ensures that requests are not lost if something happens to kafka -- the file deletion request events can be republished from the database.

Tests were slightly refactored and expanded to include testing the deletion functionality of the DAO publisher, even though that is not used at this time.


services/pcs/tests_pcs/fixtures/joint.py
```diff
@@ -22,53 +22,54 @@
...
-                yield JointFixture(
-                    config=config,
-                    file_deletion=file_deletion,
-                    rest_client=rest_client,
-                    kafka=kafka,
-                    token=token,
-                )
+    config = get_config(sources=[mongo_kafka.config, kafka.config, token_hash_config])
+    async for dao in get_file_deletion_dao(config=config):
+        async with prepare_core(config=config) as file_deletion:
```
mephenor, line 62:
```suggestion
    async with (get_file_deletion_dao(config=config) as dao, prepare_core(config=config) as file_deletion):
```
Needs some formatting, but should work the same as now and the intent with the dao is clear, the current `for` is quite confusing

services/pcs/tests_pcs/test_typical_journey.py
```diff
@@ -17,46 +17,94 @@
...
 
+pytestmark = pytest.mark.asyncio()
 
-@pytest.mark.asyncio
-async def test_journey(joint_fixture: JointFixture):  # noqa: F811
-    """Simulates a typical, successful API journey."""
-    file_id = "test_id"
 
-    files_deletion_event = event_schemas.FileDeletionRequested(file_id=file_id)
+test_file_id = "test_id"
```
mephenor, line 35:
```suggestion
TEST_FILE_ID = "test_id"
```

services/pcs/src/pcs/inject.py
```diff
@@ -21,22 +21,42 @@
...
+
+
+async def get_mongo_kafka_dao_factory(
+    config: Config,
+) -> AsyncGenerator[MongoKafkaDaoPublisherFactory, None]:
+    """Get a MongoDB DAO publisher factory."""
+    async with MongoKafkaDaoPublisherFactory.construct(config=config) as factory:
+        yield factory
+
+
```
mephenor, line 42:
```suggestion
@asynccontextmanager
```

services/pcs/src/pcs/inject.py
```diff
@@ -21,22 +21,42 @@
...
+        yield file_deletion_dao
 
 
 @asynccontextmanager
 async def prepare_core(*, config: Config) -> AsyncGenerator[FileDeletionPort, None]:
     """Construct and initialize the core component and its outbound dependencies."""
-    async with KafkaEventPublisher.construct(config=config) as event_pub_provider:
-        event_publisher = EventPubTranslator(config=config, provider=event_pub_provider)
-        file_deletion = FileDeletion(event_publisher=event_publisher)
+    async for file_deletion_dao in get_file_deletion_dao(config=config):
```
mephenor, line 58:
```suggestion
    async with get_file_deletion_dao(config=config) as file_deletion_dao:
```

services/pcs/src/pcs/adapters/outbound/daopub.py
```diff
@@ -0,0 +1,80 @@
...
+        """Configure with provider for the DaoFactoryProtocol"""
+        self._dao_publisher_factory = dao_publisher_factory
+        self._file_deletions_collection = config.file_deletions_collection
+        self._file_deletion_topic = config.files_to_delete_topic
+
+    @staticmethod
+    def _file_id_to_event(file_deletion_event: FileDeletionRequested) -> JsonObject:
+        """Translate a user to an event."""
+        validated_model = FileDeletionRequested(file_id=file_deletion_event.file_id)
+        return validated_model.model_dump()
```
mephenor, line 66: Couldn't you just directly `model_dump()` the event model instead?
Seems like unnecessary indirection

mephenor (changes requested):
A few smaller issues, with one that needs changes in a few places:
The `get_file_deletion` dao can be converted to an async context manager for its usage to be in line with the other functionality in the joint fixture

services/pcs/tests_pcs/test_typical_journey.py
TheByronHimes: Forgot about the upper case, thanks

services/pcs/src/pcs/inject.py
TheByronHimes: You're right, that is better.

services/pcs/src/pcs/adapters/outbound/daopub.py
TheByronHimes: Yes, I replaced it with a lambda. Not sure why I did that, maybe changed things around and just didn't catch it.

mephenor (approved):

PR #24, opened 2024-06-12
IFRS Outbox Subscriber (GSI-778)

TheByronHimes:
Replaces the old event subscriber with the new outbox subscriber.
There is one translator for each event type (`NonstagedFileRequestedListener`, `FileDeletionRequestedListener`, and `FileValidationSuccessListener`), which are bundled together by `KafkaOutboxSubscriber.construct()` during injection/setup. 

Where the old event subscriber called the main methods, such as `delete_file` or `stage_registered_file`, on the file registry directly, these new upsert methods sit between the subscriber and the action in the file registry. The purpose is to perform an idempotence check and store the event once the other method calls have been completed. 

### Idempotence
There is a new `IdempotenceHandler` class with 3 methods:
- `upsert_nonstaged_file_requested`
- `upsert_file_deletion_requested`
- `upsert_file_upload_validation_success`

These correspond to the events consumed by the outbox subscriber. 
There are also 3 DAOs exclusive to the `IdempotenceHandler`, one for each of the events.
When one of the aforementioned methods is called, the handler uses the corresponding DAO to see if the event is new or if it has already been processed. That logic is contained by two methods: 
- `make_record_from_update` (will produce a model that contains the update + correlation ID)
- `assert_record_is_new` (check if the event is a repeat or not and emit a debug log if not)

If the idempotence check is passed, the data is sent to the appropriate `FileRegistry` method.

For example, the following call chain from the previous code:
`translator._consume_file_downloads()` -> `FileRegistry.stage_registered_file()`
is now:
`translator.changed()` -> `IdempotenceHandler.upsert_file_upload_validation_success()` -> `FileRegistry.stage_registered_file()`

Poorly drawn representation:
image

The `IdempotenceHandler` can be overridden in the `prepare_outbox_subscriber` function for test purposes.

### Testing
The new tests are set up to check the following:
- Verify that the outbox subscriber emits warning logs if a "deleted"-type event is encountered (but otherwise does nothing)
- Verify that the correct `upsert_xyz` method is called by the outbox subscriber for a given event
- Verify that the idempotence check is carried out properly inside of each `upsert_xyz` method
- Verify that the correct processing method, such as `delete_file` or `stage_registered_file`, is called from each `upsert_xyz` method
- Verify that `assert_record_is_new` and `make_record_from_update` do what they are expected to
- Verify that the log content upon encountering duplicate records is correct


services/ifrs/src/ifrs/adapters/inbound/event_sub.py
```diff
@@ -15,138 +15,133 @@
...
+            resource_id,
         )
 
-        await self._file_registry.delete_file(file_id=validated_payload.file_id)
 
-    async def _consume_validated(
+class FileValidationSuccessListener(
+    DaoSubscriberProtocol[event_schemas.FileUploadValidationSuccess]
+):
+    """A class that consumes FileDeletionRequested events."""
```
Cito, line 120:
```suggestion
    """A class that consumes FileValidationSuccess events."""
```

services/ifrs/src/ifrs/adapters/inbound/utils.py
```diff
@@ -0,0 +1,61 @@
...
+
+def make_record_from_update(
+    record_type: type[RecordType], update: BaseModel
+) -> RecordType:
+    """Get an IdempotenceRecord containing the update payload and correlation ID."""
+    correlation_id = get_correlation_id()
+    return record_type(correlation_id=correlation_id, **update.model_dump())
+
+
+async def assert_record_is_new(
```
Cito, line 41: Consider renaming it to `check_record_is_new` or just `record_is_new`, since I expect a method called `assert_...` to return nothing, but throw an AssertionError or other kind of error in case of failure.

services/ifrs/tests_ifrs/test_outbox_subscriber.py
```diff
@@ -0,0 +1,276 @@
...
+
+TEST_FILE_DELETION_REQUESTED = event_schemas.FileDeletionRequested(file_id=TEST_FILE_ID)
+
+
+def test_make_record_from_update():
+    """Test the get_record function"""
+    record = make_record_from_update(
+        models.FileDeletionRequestedRecord, TEST_FILE_DELETION_REQUESTED
+    )
+    assert record.model_dump() == {
```
Cito, line 74: Also check that `isinstance(record, FileDeletionRequestedRecord)`

services/ifrs/tests_ifrs/test_outbox_subscriber.py
```diff
@@ -0,0 +1,276 @@
...
+        "correlation_id": get_correlation_id(),
+        "file_id": TEST_FILE_ID,
+    }
+
+
+@pytest.mark.asyncio()
```
Cito, line 80: Use `pytestmark = pytest.mark.asyncio()` so you don't need to annotate it every time.

Cito (approved):
Thanks. Very clearly documented and implemented, looks good to me.

services/ifrs/tests_ifrs/test_outbox_subscriber.py
Cito: Ok, then we should leave it like that.

services/ifrs/src/ifrs/adapters/inbound/event_sub.py
TheByronHimes: Thanks

services/ifrs/tests_ifrs/test_outbox_subscriber.py
TheByronHimes: Fixed

Cito (approved):

PR #25, opened 2024-06-13
DCS Outbox Subscriber (GSI-775)

TheByronHimes:
Download Controller Service 
---
The event `FileDeletionRequested` is now handled by the outbox subscriber instead of the normal event subscriber. 
*Both subscribers still exist, run in parallel*.

The core already handles file deletions in an idempotent manner, so no intermediary class is required.

DCS version is bumped from `1.4.0` -> `2.0.0`


services/dcs/src/dcs/adapters/inbound/event_sub.py
```diff
@@ -92,23 +93,52 @@ async def _consume_files_to_register(self, *, payload: JsonObject) -> None:
...
-            await self._consume_file_deletions(payload=payload)
         else:
             raise RuntimeError(f"Unexpected event of type: {type_}")
+
+
+class OutboxSubTranslatorConfig(BaseSettings):
+    """Config for the outbox subscriber"""
+
+    files_to_delete_topic: str = Field(
+        ...,
```
Cito, line 110:
```suggestion
        default=...,
```

Cito (approved):
Nice, looks good to me.

services/dcs/src/dcs/adapters/inbound/event_sub.py
TheByronHimes: Fixed!

Cito (approved):

PR #26, opened 2024-06-14
IRS Outbox Subscriber (GSI-779)

TheByronHimes:
Adds an outbox subscriber to handle the FileUploadReceived events. There is no extra idempotence mechanism added in this PR because that is already handled by the Interrogator class via the Fingerprint process. The `changed` method on the outbox translator class has the same action as the event sub translator when consuming the event, and calls the `interrogate` method on the interrogator. 
The `deleted` method will log a warning, as is done in similar implementations for outbox consumption of normal events.
Because there is a config change, the major version number is bumped from 2 -> 3

### DIFF
--- a/services/irs/src/irs/inject.py
+++ b/services/irs/src/irs/inject.py
@@ -32,10 +32,7 @@ from irs.adapters.inbound.event_sub import (
     EventSubTranslator,
     FileUploadReceivedSubTranslator,
 )
-from irs.adapters.outbound.dao import (
-    FingerprintDaoConstructor,
-    StagingObjectDaoConstructor,
-)
+from irs.adapters.outbound.dao import get_fingerprint_dao, get_staging_object_dao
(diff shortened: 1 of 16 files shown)


services/irs/src/irs/inject.py
```diff
@@ -92,14 +94,36 @@ async def prepare_event_subscriber(
...
+) -> AsyncGenerator[KafkaOutboxSubscriber, None]:
+    """Construct and initialize an outbox subscriber with all its dependencies.
+
+    By default, the core dependencies are automatically prepared but you can also
+    provide them using the interrogator_override parameter.
+    """
+    async with prepare_core_with_override(
+        config=config, interrogator_override=interrogator_override
+    ) as interrogator:
+        outbox_sub_translator = FileUploadReceivedSubTranslator(
```
Cito, line 111: you can use the new syntax here like you do in main.py to avoid nesting

Cito (approved):

PR #27, opened 2024-06-14
UCS Outbox Subscriber (GSI-774)

TheByronHimes:
Adds an outbox subscriber/translator with minimal changes otherwise. Idempotence is built in to the core, so no IdempotenceHandler is needed.

Previously, the UCS would publish a `FileDeletionSuccessful` event anytime it consumed a `FileDeletionRequested` event, even if nothing was there to be deleted (e.g. in the case of an re-processing or something). This has been changed so an event is only published if an upload attempt is discovered. I didn't know whether to base it on the object deletion or the upload attempts and went with the broader thing, but this can easily be changed.


Cito (approved):
Looks good to me

PR #28, opened 2024-06-17
FIS Outbox Publisher (GSI-805)

TheByronHimes:
This PR replaces the event publisher with a mongo-kafka outbox dao for the `FileUploadValidationSuccess` event (the only event published in this service). Existing tests were modified to account for the different kafka topic, and one new test was added to make sure the dto-to-event aspect of the DAO publisher provides the expected output.


services/fis/tests_fis/fixtures/joint.py
```diff
@@ -66,35 +68,38 @@ class JointFixture:
...
-                    token=token,
-                    kafka=kafka,
-                    rest_client=rest_client,
-                    s3_endpoint_alias=config.selected_storage_alias,
-                    upload_metadata_processor=upload_metadata_processor,
-                    legacy_upload_metadata_processor=legacy_upload_metadata_processor,
-                )
+        ) as app,
+    ):
+        async with AsyncTestClient(app=app) as rest_client:
```
Cito, line 95: why don't you use the new syntax here as well?

services/fis/tests_fis/test_outbox_dao.py
```diff
@@ -0,0 +1,59 @@
...
+
+        expected_event = ExpectedEvent(
+            payload=dto.model_dump(), type_="upserted", key=dto.file_id
+        )
+
+        async with joint_fixture.kafka.expect_events(
+            events=[expected_event],
+            in_topic=config.file_upload_validation_success_topic,
+        ):
+            async with set_correlation_id(new_correlation_id()):
```
Cito, line 58: again, could use the new syntax

Cito (approved):

services/fis/tests_fis/test_outbox_dao.py
TheByronHimes: as above, so here

services/fis/tests_fis/fixtures/joint.py
TheByronHimes: It is so ([5c19d57](https://github.com/ghga-de/file-services-backend/pull/28/commits/5c19d57bda7d294a16f09373d32619cb4d1a0261))

Cito (approved):

PR #30, opened 2024-06-18
Assorted cleanup

TheByronHimes:
Address an assortment of issues:
- Outdated config and lock files
- Inconsistencies and usability problems with the scripts/ dev tool
- Removed some config (a now-unused event type) that had been left in with the PCS outbox update
- Update code to python 3.12 where that had been missed


Cito (approved):
Thanks for the clean-up.

PR #31, opened 2024-06-18
PCS - Add Republish Command (GSI-813)

TheByronHimes:
Enable the republish feature of mongo-kafka DAOs with `republish-events`. This instantiate a configured DAO and run `.republish()`.
The test checks for correlation ID reinstatement and event content.

### DIFF
--- a/services/pcs/src/pcs/cli.py
+++ b/services/pcs/src/pcs/cli.py
@@ -19,7 +19,7 @@ import asyncio
 
 import typer
 
-from pcs.main import run_rest_app
+from pcs.main import republish_events, run_rest_app
 
 cli = typer.Typer()
 
@@ -28,3 +28,9 @@ cli = typer.Typer()
 def sync_run_rest_app():
     """Run the HTTP REST API."""
     asyncio.run(run_rest_app())
+
+
+@cli.command(name="republish-events")
+def sync_run_republish():
+    """Republish all outbox events regardless of publish status."""
+    asyncio.run(republish_events())
--- a/services/pcs/src/pcs/main.py
+++ b/services/pcs/src/pcs/main.py
@@ -19,7 +19,7 @@ from ghga_service_commons.api import run_server
 from hexkit.log import configure_logging
 
 from pcs.config import Config
-from pcs.inject import prepare_rest_app
+from pcs.inject import get_file_deletion_dao, prepare_rest_app
 
 
 async def run_rest_app():
@@ -29,3 +29,12 @@ async def run_rest_app():
 
     async with prepare_rest_app(config=config) as app:
         await run_server(app=app, config=config)
+
+
+async def republish_events():
+    """Republish all outbox events regardless of publish status."""
+    config = Config()
+    configure_logging(config=config)
+
+    async with get_file_deletion_dao(config=config) as dao:
+        await dao.republish()
--- a/services/pcs/tests_pcs/test_republish.py
+++ b/services/pcs/tests_pcs/test_republish.py
@@ -16,7 +16,7 @@
 """Testing for the republish functionality."""
 
 import pytest
-from ghga_event_schemas import pydantic_ as event_schemas
+from ghga_event_schemas.pydantic_ import FileDeletionRequested
 from hexkit.correlation import get_correlation_id, set_new_correlation_id
 from hexkit.protocols.daosub import DaoSubscriberProtocol
 from hexkit.providers.akafka import KafkaOutboxSubscriber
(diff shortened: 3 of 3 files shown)


services/pcs/src/pcs/cli.py
```diff
@@ -28,3 +28,9 @@
 def sync_run_rest_app():
     """Run the HTTP REST API."""
     asyncio.run(run_rest_app())
+
+
+@cli.command(name="publish-events")
+def sync_run_republish(all: bool = False):
+    """Republish all outbox events regardless of publish status."""
```
Cito, line 35: the docstring should be the same as for publish_events in main.py

services/pcs/tests_pcs/test_republish.py
```diff
@@ -0,0 +1,131 @@
...
+        self.correlation_ids = []
+        self.consumed_events = []
+
+    async def changed(self, resource_id: str, update: FileDeletionRequested) -> None:
+        """Consume event and record correlation ID"""
+        self.correlation_ids.append(get_correlation_id())
+        self.consumed_events.append(resource_id)
+
+    async def deleted(self, resource_id: str) -> None:
+        """Dummy"""
```
Cito, line 51: Maybe raise an error in this method to verify that it is not called.

services/pcs/tests_pcs/test_republish.py
```diff
@@ -0,0 +1,131 @@
...
+        document = dto_to_document(unpublished_event, id_field="file_id")
+        collection.insert_one(document)
+
+    expected_unpublished = ExpectedEvent(
+        payload=unpublished_event.model_dump(),
+        type_="upserted",
+        key="unpublished_event",
+    )
+
+    # Verify that the only the unpublished event is published
```
Cito, line 85:
```suggestion
    # Verify that only the unpublished event is published
```

services/pcs/tests_pcs/test_republish.py
```diff
@@ -0,0 +1,131 @@
...
+pytestmark = pytest.mark.asyncio()
+
+
+class DummySubTranslator(DaoSubscriberProtocol):
+    """A class that consumes FileDeletionRequested events."""
+
+    event_topic: str = ""
+    dto_model = FileDeletionRequested
+    correlation_ids: list[str]
+    consumed_events: list[str]
```
Cito, line 38: I would use a single list with two-tuples here.

Cito (changes requested):
Looks good, some minor comments.

services/pcs/src/pcs/cli.py
TheByronHimes: Makes sense, made it so

services/pcs/tests_pcs/test_republish.py
TheByronHimes: Good idea

services/pcs/tests_pcs/test_republish.py
TheByronHimes: Nice, I have done that in [e510694](https://github.com/ghga-de/file-services-backend/pull/31/commits/e5106941e6db6cfcfb129857acef794ff9353431)

services/pcs/src/pcs/cli.py
```diff
@@ -28,3 +28,9 @@
 def sync_run_rest_app():
     """Run the HTTP REST API."""
     asyncio.run(run_rest_app())
+
+
+@cli.command(name="publish-events")
+def sync_run_publish_events(all: bool = False):
+    """Publish pending events. Use `--all` to (re)publish all events regardless of status."""
```
Cito, line 35:
```suggestion
def sync_run_publish_events(all: Annotated[bool, typer.Argument(help="to (re)publish all events regardless of status")] = False):
    """Publish pending events."""
```

Maybe this is even better.

services/pcs/src/pcs/main.py
```diff
@@ -29,3 +29,15 @@ async def run_rest_app():
 
     async with prepare_rest_app(config=config) as app:
         await run_server(app=app, config=config)
+
+
+async def publish_events(*, all: bool = False):
+    """Publish pending events. Use `--all` to (re)publish all events regardless of status."""
```
Cito, line 35:
```suggestion
async def publish_events(*, all: bool = False):
    """Publish pending events. Set `all` to (re)publish all events regardless of status."""
```

Cito: Since this docstring is not shown in the CLI.

Cito (approved):
One more idea for the comments.

Not really important, but this code will be probably copied many times, so worth some polishing.

PR #33, opened 2024-06-20
DCS Outbox Publisher (GSI-772)

TheByronHimes:
Adds the outbox publisher for the NonStagedFileRequested event, as well as the `publish-events [--all]` command.

### DIFF
--- a/services/dcs/src/dcs/adapters/outbound/daopub.py
+++ b/services/dcs/src/dcs/adapters/outbound/daopub.py
@@ -14,7 +14,7 @@
 # limitations under the License.
 """Adapter for publishing outbox events to other services."""
 
-from ghga_event_schemas.pydantic_ import FileUploadValidationSuccess
+from ghga_event_schemas.pydantic_ import NonStagedFileRequested
 from hexkit.protocols.daopub import DaoPublisher, DaoPublisherFactoryProtocol
 from pydantic import Field
 from pydantic_settings import BaseSettings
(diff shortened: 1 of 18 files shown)


services/dcs/src/dcs/adapters/outbound/daopub.py
```diff
@@ -0,0 +1,74 @@
...
+        ),
+        examples=["file_downloads"],
+    )
+    unstaged_download_collection: str = Field(
+        default=...,
+        description=(
+            "The type used for event indicating that a download was requested"
+            + " for a file that is not yet available in the outbox."
+        ),
+        examples=["unstaged_download_requested"],
```
mephenor, line 42: Not sure if this has to be hyphenated instead

services/dcs/tests_dcs/test_outbox_dao.py
```diff
@@ -0,0 +1,163 @@
...
+    async with set_new_correlation_id():
+        await joint_fixture.nonstaged_file_requested_dao.republish()
+
+        # consume the republished events with the dummy translator
+        translator = DummySubTranslator(config=joint_fixture.config)
+        async with KafkaOutboxSubscriber.construct(
+            config=joint_fixture.config, translators=[translator]
+        ) as subscriber:
+            await subscriber.run(forever=False)
+            await subscriber.run(forever=False)
```
mephenor, line 158: As in the other PR: wouldn't this just consume the initial insert events?

TheByronHimes: You're right, I misunderstood what happened in the expect_events cm. I've updated it to clear the topics right after that so the only events consumed are the republished ones.

services/dcs/src/dcs/adapters/outbound/daopub.py
TheByronHimes: It does, I've updated the description.

mephenor (approved):

PR #35, opened 2024-06-21
pcs and fis: Fix publish-events command

TheByronHimes:
When `all` is annotated with `typer.Argument`, it doesn't work like a flag very well. Using `typer.Option` gives it the intuitive behavior, i.e. `fis publish-events` or `fis publish-events --all`. Using typer.Argument and `--all` results in an error.


Cito (approved):
You're right, this is actually an option, not an argument.

PR #42, opened 2024-07-03
Adapt to DS-Kit storage selection

mephenor:
Changes included in this PR:

- The `selected_storage_alias` config option has been removed. Instead this is now part of both the legacy and non-legacy payload
- The non-legacy, `federated/file_ingest` endpoint no longer expects an encrypted payload and decryption functionality for this path has been removed.

The initial draft of the epic docs called for server-side verification of the storage alias, but this is not included, as the DS-Kit already get's it's necessary information from the WKVS and does not progress through initialization, if the given `storage_alias` that becomes part of the payload is unknown.


lkuchenb (approved):
LGTM, thank you!

PR #45, opened 2024-07-10
Docker workflow improvements

mephenor:
Found an easy solution to check if a docker image already exists:
Just pull the image, and if successful, fail the step by exiting with a non-zero code.
If pulling fails, just do nothing.
Needs an `or` in after the pull command, else it fails on an unsuccessful pull.

This also renames the action and removes the suffix sha from the docker tag.

### DIFF
--- a/.github/workflows/docker_on_release.yaml
+++ b/.github/workflows/docker_on_release.yaml
@@ -156,6 +156,22 @@ jobs:
         run: |
           echo version=$(python3.12 -c "import tomllib; print(tomllib.load(open('services/${{ matrix.service }}/pyproject.toml', 'rb'))['project']['version'])") >> $GITHUB_OUTPUT
 
+      - name: Create current docker tag
+        id: docker-tag
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/docker_on_release.yaml
```diff
@@ -156,6 +156,22 @@ jobs:
...
 
+      - name: Create current docker tag
+        id: docker-tag
+        run: |
+          echo tag="${{ env.DOCKERHUB_NAMESPACE }}/${{ steps.extract-service-name.outputs.name }}:${{ steps.extract-service-version.outputs.version }}" >> $GITHUB_OUTPUT
+
+      - name: Check if docker image already exists
+        id: check-image-exists
+        run: |
+          # pipefail set by default, join with or to not cancel run on missing image
```
Cito, line 167:
```suggestion
          # pipefail set by default, join with "or" to not cancel run on missing image
```

Cito: can be confusing to read without the quotes in mind :)

Cito (approved):
Great you could find such a simple solution.

PR #47, opened 2024-07-17
General cleanup

mephenor:
- Renamed test workflow to be picked up by readme again & simplified name for badge
- Updated `check_license` and `update_lock` scripts - with a tiny bit of custom stuff in check_license
- Updated lock files
- Added py.typed marker to each service
- Changed service pyproject.tomls repository urls to point to the respective subdirectory & brought the setuptools part in line
- Changed root readme to point to service folders instead of directly to the service readmes - those are displayed directly below in the browser either way


Cito (approved):
Looks good to me. Nice cleanup.

PR #48, opened 2024-07-17
Added auth mount point to vault adapter

mephenor:
Exposes kubernetes adapter mount point as config option called `vault_kube_mount_point` analogously to the already existing `vault_kube_role`.


Cito (approved):
Looks good to me.

The accumulated code has become a bit ugly and inefficient. For instance, the jwt could be read only once instead of for every login. We could prepare login `kwargs` once with the parameters needed for the login and then pass them as `**kwargs` when calling login. Something for our refactoring list.

PR #50, opened 2024-07-29
Use upsert for triggering outbox event instead of insert

TheByronHimes:


Cito (approved):
Looks like the right fix.

PR #57, opened 2024-08-13
Fix Docker Hub CD to publish to correct repository

lkuchenb:


lkuchenb:
Requires additional fixes, deferred for template refactoring.

PR #62, opened 2024-11-04
Idempotency guards

mephenor:
This PR adds two simple lookups to the File Ingest Service, one for each ingest endpoint to abort any further processing if the file ID has already been processed.

As the entry in the outbox DAO is only created once everything else is done, we can reasonably assume that the presence of a corresponding entry means the ingest of a specific file has succeeded previously - at least inside the FIS.
This will still produce additional new secret entries in vault for the legacy path if something should crash before an entry is created, but the information is only propagated once in this case.

Also, decided to return to 202 for the no-op as well, as we don't need to distinguish between the new and already seen case on the client side and this would require a small change to the DS-Kit if we go with any other status code.

### DIFF
--- a/services/fis/src/fis/adapters/inbound/fastapi_/routes.py
+++ b/services/fis/src/fis/adapters/inbound/fastapi_/routes.py
@@ -17,6 +17,7 @@
 from typing import Annotated
 
 from fastapi import APIRouter, HTTPException, Response, status
+from fastapi.responses import JSONResponse
 
 from fis.adapters.inbound.fastapi_ import dummies
 from fis.adapters.inbound.fastapi_.http_authorization import (
@@ -67,10 +68,14 @@ async def ingest_legacy_metadata(
     except (DecryptionError, WrongDecryptedFormatError) as error:
         raise HTTPException(status_code=422, detail=str(error)) from error
 
-    if await upload_metadata_processor.has_already_been_processed(
-        file_id=decrypted_metadata.file_id
-    ):
-        return Response(status_code=204)
+    file_id = decrypted_metadata.file_id
+    if await upload_metadata_processor.has_already_been_processed(file_id=file_id):
+        return JSONResponse(
+            status_code=409,
+            content={
+                "error": f"Metadata for file {file_id} has already been processed."
+            },
+        )
 
     file_secret = decrypted_metadata.file_secret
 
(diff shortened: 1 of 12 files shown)


TheByronHimes (approved):
Looks good from what I can see

lkuchenb (changes requested):
@mephenor I would say we need to distinguish success from no-op and als reflect that in the client, everything else will be quite hard to maintain downstream. How does dskit currently handle individual failures? Does it log and continue with the next item or abort? Either way, let's distinguish here and then adjust dskit accordingly.

mephenor:
@lkuchenb 
In batch mode individual failures are logged and then the process continues.
I'll change the code to return 204 for both endpoints if the payload has already been processed.
This will require a one line change in DS-Kit.

services/fis/tests_fis/test_ingest.py
```diff
@@ -33,14 +34,14 @@
 
 async def test_legacy_decryption_happy(joint_fixture: JointFixture):
     """Test decryption with valid keypair and correct file upload metadata format."""
-    payload = LegacyUploadMetadata(
-        **TEST_PAYLOAD.model_dump(),
-        file_secret=base64.b64encode(os.urandom(32)).decode("utf-8"),
-    )
+    # Can't use the LegacyUploadMetadata directly, as dump_json will obfuscate the secret
```
mephenor, line 37:
```suggestion
    # Can't use LegacyUploadMetadata directly, as dump_json will obfuscate the secret
```

mephenor:
Changes to response status code as detailed above.
Also opted to change internal representation of `file_secret` to `SecretStr` and only get the value at the boundary to vault.
Should help with not accidentally logging sensitive information.

services/fis/src/fis/adapters/inbound/fastapi_/routes.py
```diff
@@ -120,9 +149,12 @@ async def ingest_secret(
...
     """Decrypt payload and deposit file secret in exchange for a secret id"""
-    file_secret = await upload_metadata_processor.decrypt_secret(
-        encrypted=encrypted_payload
-    )
+    try:
+        file_secret = await upload_metadata_processor.decrypt_secret(
+            encrypted=encrypted_payload
+        )
+    except DecryptionError as error:
+        raise HTTPException(status_code=422, detail=str(error)) from error
```
lkuchenb, line 157: Should be documented in the FastAPI header

lkuchenb (approved):
Thanks, minor remark below

lkuchenb:
@mephenor also - should this be a major version bump due to the change of API behavior?

services/fis/src/fis/adapters/inbound/fastapi_/routes.py
mephenor: 422 should be documented by default, but once I find where to get the HTTPValidationError model from, I can adjust the description.

mephenor: Added in cf8cdd1a079e0b857e9437ff57bd175790144104

lkuchenb (approved):

PR #66, opened 2024-11-12
Fix for docker image check

mephenor:


lkuchenb (commented):
I'm not sure what exactly this fixes. And why not simply `if docker pull...`?

mephenor:
@lkuchenb 
Sorry, provided the context elsewhere, but forgot it here:
The current version does not work correctly, stupid mistake, so it short circuits in the case that the pull was successful and the check passes in either case.
This should fix that behaviour.

If would be easier, but somehow haven't thought about that.
Changes would need another round of testing though, which would be a bit involved because this doesn't fully work with act locally, meaning creating another fork and testing there.

ckaipf:
LGTM now, I could successfully push to my [repository](https://hub.docker.com/repository/docker/camillkaipf/). I interrupted it during the process. When re-running the workflow it skipped the already published tags. However I got this:

```
 network="container:act-Push-to-docker-on-new-release-push-to-docker-6-d7833a195f498b37f8c41a6124c8fd5fd2e1fea19d88a6e8719c985e770c9c68"
| Running trivy with options: trivy image  --format table --exit-code  1 --ignore-unfixed --vuln-type  os,library  docker.io/***/internal-file-registry-service:2.0.3
| Global options:  
| 2024-11-13T16:20:16.208Z      INFO    Need to update DB
| 2024-11-13T16:20:16.209Z      INFO    DB Repository: ghcr.io/aquasecurity/trivy-db
| 2024-11-13T16:20:16.209Z      INFO    Downloading DB...
| 2024-11-13T16:20:17.696Z      FATAL   init error: DB error: failed to download vulnerability DB: database download error: OCI repository error: 1 error occurred:
|       * GET https://ghcr.io/v2/aquasecurity/trivy-db/manifests/2: TOOMANYREQUESTS: retry-after: 668.674µs, allowed: 44000/minute
```

We might run into this again.

ckaipf (approved):

lkuchenb (approved):

PR #68, opened 2024-11-14
Specify missing property in GitHub action

dontseyit:
This PR adds missing property to the GitHub action template. 

The failing action run: [Required property is missing: shell](https://github.com/ghga-de/file-services-backend/actions/runs/11839884234/job/32992461894)

Apparently when using composite actions we have to specify the shell in each step. Related documentation: https://docs.github.com/en/actions/sharing-automations/creating-actions/creating-a-composite-action#creating-an-action-metadata-file

### DIFF
--- a/.github/workflows/push_to_docker/action.yaml
+++ b/.github/workflows/push_to_docker/action.yaml
@@ -51,10 +51,12 @@ runs:
 
     - name: Symlink requirement files
       id: symlink-requirements
+      shell: bash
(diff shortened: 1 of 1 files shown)


mephenor:
@dontseyit Sorry, didn't see your PR and opened another one: https://github.com/ghga-de/file-services-backend/pull/69
Can you approve that one instead, that one has a few more changes.

Apparently act does not replicate the correct behaviour here wrt the missing `shell:bash`

PR #76, opened 2024-12-10
Server keypair validators and decoding error handling

mephenor:
- Both the server public and private key are now decoded and the length is checked in a validator to catch issues early and crash on startup instead of on the first request.
- An additional Response, HttpDecodingError, is now produced by both endpoints receiving base64 encoded data. Failure to decode either the client/requester public key or the file part means something went wrong before the data hits the EKSS and accordingly should be handled by the caller instead of crashing the EKSS

### DIFF
--- a/services/ekss/src/ekss/adapters/inbound/fastapi_/router.py
+++ b/services/ekss/src/ekss/adapters/inbound/fastapi_/router.py
@@ -58,7 +58,7 @@ ERROR_RESPONSES = {
     },
     "decodingError": {
         "description": (
-            "One of the inputs provided to the API could not be decoded as base64 string."
+            "One of the provided inputs could not be decoded as base64 string."
         ),
         "model": exceptions.HttpDecodingError.get_body_model(),
     },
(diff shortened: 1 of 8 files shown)


services/ekss/src/ekss/adapters/inbound/fastapi_/router.py
```diff
@@ -33,25 +33,35 @@
...
     "envelopeDecryptionError": {
-        "description": (""),
+        "description": (
+            "Could not decrypt Crypt4GH header envelope with the given key pair."
+        ),
         "model": exceptions.HttpEnvelopeDecryptionError.get_body_model(),
     },
     "secretInsertionError": {
-        "description": (""),
+        "description": ("Failed to successfully inset secret into vault."),
```
lkuchenb, line 48:
```suggestion
        "description": ("Failed to successfully insert secret into vault."),
```

services/ekss/src/ekss/adapters/inbound/fastapi_/router.py
```diff
@@ -33,25 +33,35 @@
...
         "model": exceptions.HttpVaultConnectionError.get_body_model(),
     },
     "secretNotFoundError": {
-        "description": (""),
+        "description": ("Could not find a secret for the given secret ID."),
         "model": exceptions.HttpSecretNotFoundError.get_body_model(),
     },
+    "decodingError": {
+        "description": (
+            "One of the inputs provided to the API could not be decoded as base64 string."
```
lkuchenb, line 61:
```suggestion
            "One of the provided inputs could not be decoded as base64 string."
```

lkuchenb (changes requested):
Thanks, minor requests below

lkuchenb (approved):

PR #81, opened 2025-01-14
Simplify ekss test setup code

TheByronHimes:


mephenor (approved):
Thanks, that's easier to read

PR #85, opened 2025-01-20
IFRS: prototype migration framework and object_size migration (GSI-1271)

TheByronHimes:
Adds a very rough version of the migration framework which aims to minimize the amount of related code required to be maintained across services without being too restricting/prescriptive.

**Step 1**: The logic for every migration should subclass `MigrationDefinition` and, usually, `Reversible` as well:
```python
class V2Migration(MigrationDefinition, Reversible):
    """Adds `object_size` and migrates the database to version 2"""

    version = 2

    async def apply(self):
        async with self.auto_finalize("file_metadata"):
            await self.migrate_docs_in_collection(
                coll_name=METADATA_COLLECTION,
                change_function=my_change_function,
                validation_model=FileMetadata,
                id_field="file_id",
            )

    async def unapply(self):
        ...
```

**Step 2**: Define the current database version and the migration map. 
>In this prototype, version 0 = "DB versioning is not set up" and version 1 = "DB versioning has been created", and our migrations begin with version 2. Does not have to be that way, that's just what I landed on for now.
```python
from my_migrations import V2Migration, V3Migration, V4Migration  # etc.

MIGRATION_MAP = {2: V2Migration, 3: V3Migration, 4: V4Migration}
DB_VERSION = 4
```

**Step 3**: Instantiate the `MigrationManager` via `async with`, then call `.migrate_or_wait()`:
```python
MIGRATION_MAP = {2: V2Migration, 3: V3Migration, 4: V4Migration}
DB_VERSION = 4

async def run_db_migrations(*, config: Config, target_version: int = DB_VERSION):
    """Check if the database is up to date and attempt to migrate the data if not."""
    async with MigrationManager(
        config=config,
        target_version=target_version,
        migration_map=MIGRATION_MAP,
    ) as mm:
        await mm.migrate_or_wait()
```

**Step 4**: Perform the DB migrations/version check sometime before firing up the actual service code:
```python
async def consume_events(run_forever: bool = True):
    """Run an event consumer listening to the specified topics."""
    config = Config()
    await run_db_migrations(config=config)

    async with prepare_outbox_subscriber(config=config) as outbox_subscriber:
        await outbox_subscriber.run(forever=run_forever)
```

### Notes
The `MigrationDefinition` class defines some useful logic for common operations, such as "staging" migrated data (where we rename the collections to swap the new data with the old), dropping new collections, generating temp table prefixing, applying an update function to each doc in a collection with validation, and a context manager to automate staging/dropping/error case cleanup.

The `MigrationManager` handles higher-level responsibilities and is ignorant of migration implementation details. It cares about whether or not the database is set up for migrations, determining which migrations it needs to run to get from version X to version Y, and running said migrations. It handles the database "lock" document to ensure only one service instance ever performs migrations, as well as maintaining a record past migrations.

Solving the catch-22 of making sure only one instance _initializes_ the lock & versioning collections is straightforward: the code inserts a lock document with a fixed ID. If the result is a `DuplicateKeyError` error, stop and enter the waiting cycle (another instance was faster), otherwise continue with the init & migration logic. The service instance that performs the initialization will already have the lock acquired after finishing setup and can roll right into executing any migrations.

Currently one test case added that covers the following:
- Migration success
- Migration unapply success
- Migration unapply failure due to model mismatch
- Migration idempotence (should be able to run migration check a bunch with no effect if already up to date)

Outstanding items planned or in progress for official framework:
- batch processing (naive one-by-one doc updates right now)
- index copying (need to understand this better)
- (maybe) automatically pass config to instances of MigrationDefinition - only an inconvenience for testing though
- Testing to cover all the edge cases along with unit testing to better catch problems with helper functions
- Better metrics (e.g. # documents affected)
- Logging/error handling is currently unpolished, needs improvement
- Limits or something for retries? Not sure
- `Reversible` subclass thing could probably be done differently
- Better documentation
- Whatever ideas and critiques come up in the meantime

### DIFF
--- a/services/ifrs/src/ifrs/migration_logic/_manager.py
+++ b/services/ifrs/src/ifrs/migration_logic/_manager.py
@@ -1,0 +1,420 @@
+# Copyright 2021 - 2024 Universität Tübingen, DKFZ, EMBL, and Universität zu Köln
+# for the German Human Genome-Phenome Archive (GHGA)
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Tools to run database migrations in services"""
+
+import logging
+from contextlib import asynccontextmanager
+from time import sleep, time
+from typing import Any, Literal
+
+from ghga_service_commons.utils.utc_dates import now_as_utc
+from hexkit.providers.mongodb import MongoDbConfig
+from motor.motor_asyncio import AsyncIOMotorClient, AsyncIOMotorDatabase
+from pydantic import BaseModel, Field
+from pymongo.errors import DuplicateKeyError
+
+from ._utils import MigrationDefinition, Reversible
+
+log = logging.getLogger(__name__)
+
+MigrationType = Literal["FORWARD", "BACKWARD"]
+MigrationCls = type[MigrationDefinition]
+MigrationMap = dict[int, MigrationCls]
+
+
+def duration_in_ms(duration: float) -> int:
+    return int(duration * 1000)
+
+
+class MigrationConfig(MongoDbConfig):
+    """Minimal configuration required to run the migration process."""
+
+    lock_collection: str = Field(
+        ...,
+        description="The name of the collection containing the DB Lock document for this service",
+        examples=["ifrsLock", "dcsLock"],
+    )
+    db_version_collection: str = Field(
+        ...,
+        description="The name of the collection containing DB version information for this service",
+        examples=["ifrsDbVersions"],
+    )
+    migration_wait_sec: int = Field(
+        ...,
+        description="The number of seconds to wait before checking the DB version again",
+        examples=[5, 30, 180],
+    )
+
+
+class DbVersionRecord(BaseModel):
+    """Model containing information about DB versions and how they were achieved."""
+
+    version: int = Field(..., description="The database version")
+    details: dict[str, Any] = Field(
+        ...,
+        description="Extra information about the migration to this version. ",
+        examples=[
+            {
+                "completed": "2025-01-17T13:42:58.396538+00:00",
+                "migration_type": "FORWARD",
+                "total_duration_ms": 5000,
+            }
+        ],
+    )
+
+
+class MigrationStepError(RuntimeError):
+    """Raised when a specific migration step fails, e.g. migrating from v4 to v5"""
+
+    def __init__(self, *, current_ver: int, target_ver: int, err_info: str):
+        msg = (
+            f"Unable to migrate from DB version {current_ver} to {target_ver}."
+            + f" Cause:\n  '{err_info}'"
+        )
+        super().__init__(msg)
+
+
+class DbLockError(RuntimeError):
+    """Raised when the DB lock can't be released or acquired due to an error."""
+
+    def __init__(
+        self, *, op: Literal["acquire", "release"], coll_name: str, err_info: str
+    ):
+        msg = (
+            f"Failed to {op} the lock in DB lock collection {coll_name}."
+            + f" Error details:\n  '{err_info}'"
+        )
+        super().__init__(msg)
+
+
+class DbVersioningInitError(RuntimeError):
+    """Raised when DB versioning initialization fails due to an error."""
+
+    def __init__(self, *, err_info: str):
+        msg = f"DB versioning initialization failed. Error details:\n  {err_info}"
+        super().__init__(msg)
+
+
+def _get_db_version_from_records(version_docs: list[DbVersionRecord]) -> int:
+    """Gets the current DB version from the documents found in the version collection."""
+    # Make sure we know what the latest version is, not just the max
+    version_docs.sort(key=lambda doc: doc.details["completed"])
+    return version_docs[-1].version if version_docs else 0
+
+
+class MigrationManager:
+    """Top-level logic for ensuring the database is updated before running the service.
+
+    The `migrate_or_wait` method must be called before any instance of the service
+    begins its main execution loop.
+
+    Reserved DB version numbers:
+    - 0 = db versioning not yet implemented
+    - 1 = db versioning implemented
+
+    Example usage:
+    ```
+    from my_service.config import Config  # inherits from MongoDbConfig
+    from ghga_service_commons.migration import MigrationManager, MigrationMap
+
+    DB_VERSION = 2  # the current expected DB version
+    MY_MIGRATION_MAP = MigrationMap({
+        2: V2Migration,
+        # future migrations will go here
+    })
+
+    def migrate_my_service():
+        # Called before starting my_service
+        config = Config()
+
+        with MigrationManager(config, DB_VERSION, MY_MIGRATION_MAP) as mm:
+            mm.migrate_or_wait()
+    ```
+    """
+
+    client: AsyncIOMotorClient
+    db: AsyncIOMotorDatabase
+
+    def __init__(
+        self,
+        config: MigrationConfig,
+        target_version: int,
+        migration_map: MigrationMap,
+    ):
+        """Instantiate the MigrationManager.
+
+        Args
+        - `config`: Config containing db connection str and lock/db versioning collections
+        - `target_version`: Which version the db needs to be at for this version of the service
+        - `migration_map`: A dict with the MigrationDefinition class for each db version
+        """
+        if target_version < 1:
+            raise RuntimeError("Expected database version must be 1 or greater")
+
+        self.config = config
+        self.target_ver = target_version
+        self.migration_map = migration_map
+        self._lock_acquired = False
+        self._entered = False
+        self._migration_type: MigrationType = "FORWARD"
+
+    async def __aenter__(self):
+        """Set up database client and database reference"""
+        self.client = AsyncIOMotorClient(
+            self.config.db_connection_str.get_secret_value()
+        )
+        self.db = self.client[self.config.db_name]
+        self._entered = True
+        return self
+
+    async def __aexit__(self, exc_type_, exc_value, exc_tb):
+        """Release DB lock and close/remove database client"""
+        await self._release_db_lock()
+        self.client.close()
+
+    async def _get_version_docs(self) -> list[DbVersionRecord]:
+        """Gets the DB version information from the database."""
+        collection = self.db[self.config.db_version_collection]
+        version_docs = [DbVersionRecord(**doc) async for doc in collection.find()]
+
+        return version_docs
+
+    @asynccontextmanager
+    async def _lock_db(self):
+        await self._acquire_db_lock()
+        try:
+            yield
+        finally:
+            await self._release_db_lock()
+
+    async def _acquire_db_lock(self):
+        """Try to acquire the lock on the DB and return the result.
+
+        Logs and raises any error that occurs while updating the lock document.
+        """
+        if self._lock_acquired:
+            log.debug("Database lock already acquired")
+            return
+
+        try:
+            lock_col = self.db[self.config.lock_collection]
+            lock_acquired = await lock_col.find_one_and_update(
+                {"lock_acquired": False},
+                {
+                    "$set": {
+                        "lock_acquired": True,
+                        "acquired_at": now_as_utc().isoformat(),
+                    }
+                },
+            )
+        except BaseException as exc:
+            error = DbLockError(
+                op="acquire", coll_name=self.config.lock_collection, err_info=str(exc)
+            )
+            log.error(error)
+            raise error from exc
+
+        self._lock_acquired = bool(lock_acquired)
+        if self._lock_acquired:
+            log.info("Database lock acquired")
+
+    async def _release_db_lock(self) -> None:
+        """Release the DB lock.
+
+        Logs and re-raises any errors that occur during the update.
+        """
+        if not self._lock_acquired:
+            log.debug("Database lock already released")
+            return
+        try:
+            lock_col = self.db[self.config.lock_collection]
+            await lock_col.find_one_and_update(
+                {"lock_acquired": True},
+                {"$set": {"lock_acquired": False, "acquired_at": ""}},
+            )
+            self._lock_acquired = False
+        except BaseException as exc:
+            error = DbLockError(
+                op="release",
+                coll_name=self.config.lock_collection,
+                err_info=str(exc),
+            )
+            log.critical(error)
+            raise error from exc
+        log.info("Database lock released")
+
+    async def _record_migration(self, *, version: int, total_duration_ms: int):
+        """Insert a DbVersionRecord with processing information"""
+        details: dict[str, Any] = {
+            "completed": now_as_utc().isoformat(),
+            "total_duration_ms": total_duration_ms,
+            "migration_type": self._migration_type,
+        }
+        record = DbVersionRecord(version=version, details=details)
+        version_collection = self.db[self.config.db_version_collection]
+        await version_collection.insert_one(record.model_dump())
+
+    async def _initialize_versioning(self) -> bool:
+        """Create and acquire the DB lock, then add the versioning collection.
+
+        Returns `True` if setup was performed, else `False`.
+        """
+        init_start = time()
+        lock_collection = self.db[self.config.lock_collection]
+        lock_doc = [_ async for _ in lock_collection.find()]
+        if not lock_doc:
+            # lock document has not been created yet, so add it
+            try:
+                await lock_collection.insert_one(
+                    {
+                        "_id": 0,
+                        "lock_acquired": False,
+                        "acquired_at": "",
+                    }
+                )
+            except DuplicateKeyError:
+                # another instance inserted the doc first, so stop and wait to retry
+                return False
+
+        # Lock database so other instances can't attempt migrations
+        async with self._lock_db():
+            if not self._lock_acquired:
+                return False
+
+            # Initialize db version collection
+            await self._record_migration(
+                version=1,
+                total_duration_ms=duration_in_ms(time() - init_start),
+            )
+        return True
+
+    def _get_sequence(self, *, current_ver: int, target_ver: int) -> list[int]:
+        """Return an ordered list of the version migrations to apply/unapply"""
+        # In forward case, we don't need to apply current ver
+        # in backward case, we don't want to unapply the target ver
+        step_range = (
+            range(current_ver, target_ver, -1)
+            if self._migration_type == "BACKWARD"
+            else range(current_ver + 1, target_ver + 1)
+        )
+        steps = list(step_range)
+        return steps
+
+    def _fetch_migration_cls(self, version: int) -> MigrationCls:
+        """Return the stored migration for the specified version.
+
+        Raise an error if the  doesn't exist or doesn't implement unapply when needed.
+        """
+        try:
+            migration_cls = self.migration_map[version]
+            if self._migration_type == "BACKWARD" and not issubclass(
+                migration_cls, Reversible
+            ):
+                raise RuntimeError(
+                    f"Planning to unapply migration v{version}, but"
+                    + f" it doesn't subclass `{Reversible.__name__}`!"
+                )
+            return migration_cls
+        except KeyError as err:
+            mig_type = self._migration_type.lower()
+            raise NotImplementedError(
+                f"No {mig_type} migration implemented for version {version}"
+            ) from err
+
+    async def _perform_migrations(self, *, current_ver: int):
+        """Migrate forward or backward to reach target DB version.
+
+        Raises `MigrationError` if unsuccessful.
+        """
+        seq = self._get_sequence(current_ver=current_ver, target_ver=self.target_ver)
+        migrations = [self._fetch_migration_cls(ver) for ver in seq]
+        unapplying = self._migration_type == "BACKWARD"
+        try:
+            # Execute & time each migration in order to get to the target DB version
+            for v, migration_cls in zip(seq, migrations, strict=True):
+                # Determine if this is the last migration to apply/unapply
+                last_ver_called = self.target_ver + 1 if unapplying else self.target_ver
+                is_final_migration = v == last_ver_called
+
+                # instantiate MigrationDefinition
+                migration = migration_cls(
+                    db=self.db,
+                    unapplying=unapplying,
+                    is_final_migration=is_final_migration,
+                )
+
+                # Call apply/unapply based on migration type
+                await migration.unapply() if unapplying else await migration.apply()
+        except BaseException as exc:
+            error = MigrationStepError(
+                current_ver=current_ver, target_ver=self.target_ver, err_info=str(exc)
+            )
+            log.critical(error)
+            raise error from exc
+
+    async def _migrate_db(self) -> bool | None:
+        """Ensure the database is up to date before running the actual app.
+
+        If the database is already up to date, no changes are made. If the database is
+        out of date, migration code is executed to make the database current.
+
+        Returns True if migrations are finished or up-to-date and False otherwise.
+        """
+        version_docs = await self._get_version_docs()
+        version = _get_db_version_from_records(version_docs)
+
+        if version == 0:
+            try:
+                init_complete = await self._initialize_versioning()
+            except BaseException as exc:
+                error = DbVersioningInitError(err_info=str(exc))
+                log.critical(error)
+                raise error from exc
+            if not init_complete:
+                return False
+            version = 1
+
+        if version == self.target_ver:
+            # DB is up to date, run service
+            return True
+
+        # DB version is not what it should be: acquire lock and migrate
+        async with self._lock_db():
+            if not self._lock_acquired:
+                return False
+
+            self._migration_type = (
+                "FORWARD" if version < self.target_ver else "BACKWARD"
+            )
+
+            start = time()
+            await self._perform_migrations(current_ver=version)
+            duration_ms = duration_in_ms(time() - start)
+
+            # record the db version
+            await self._record_migration(
+                version=self.target_ver,
+                total_duration_ms=duration_ms,
+            )
+        return True
+
+    async def migrate_or_wait(self):
+        """Try to migrate the database or wait until migrations are completed."""
+        if not self._entered:
+            raise RuntimeError("MigrationManager must be used as a context manager")
+
+        # need to implement some kind of total time limit, warning logging, etc. later
+        while not await self._migrate_db():
+            sleep(self.config.migration_wait_sec)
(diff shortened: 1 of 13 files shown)


services/ifrs/src/ifrs/migration_logic/_manager.py
```diff
@@ -0,0 +1,418 @@
...
+        description="The number of seconds to wait before checking the DB version again",
+        examples=[5, 30, 180],
+    )
+
+
+class DbVersionRecord(BaseModel):
+    """Model containing information about DB versions and how they were achieved."""
+
+    version: int = Field(..., description="The database version")
+    details: dict[str, Any] = Field(
```
Cito, line 65: Should this be a typed dict or a class? Why not flatten this model?

services/ifrs/src/ifrs/migration_logic/_manager.py
```diff
@@ -0,0 +1,418 @@
...
+    async def _record_migration(self, *, version: int, total_duration_ms: int):
+        """Insert a DbVersionRecord with processing information"""
+        details: dict[str, Any] = {
+            "completed": now_as_utc().isoformat(),
+            "total_duration_ms": total_duration_ms,
+            "migration_type": self._migration_type,
+        }
+        record = DbVersionRecord(version=version, details=details)
+        version_collection = self.db[self.config.db_version_collection]
+        await version_collection.insert_one(record.model_dump())
```
Cito, line 267: should we set `_id` to version here so that we can look up version records by id more easily? otherwise it will get a random object id.

services/ifrs/src/ifrs/migration_logic/_manager.py
```diff
@@ -0,0 +1,418 @@
...
+    Example usage:
+    ```
+    from my_service.config import Config  # inherits from MongoDbConfig
+    from ghga_service_commons.migration import MigrationManager, MigrationMap
+
+    DB_VERSION = 2  # the current expected DB version
+    MY_MIGRATION_MAP = MigrationMap({
+        2: V2Migration,
+        # future migrations will go here
+    })
```
Cito, line 136: Should we use configuration by convention here? E.g. if al the migration classes are called `VxMigration`, we could `import *`  them and then create the map automatically from their names.

services/ifrs/src/ifrs/migration_logic/_utils.py
```diff
@@ -0,0 +1,210 @@
...
+
+Document = dict[str, Any]
+
+
+def validate_doc(doc: Document, *, model: type[BaseModel], id_field: str):
+    """Ensure that new content passes model validation.
+
+    Also check that `dto_to_document` results in the same document.
+    """
+    as_model = document_to_dto(deepcopy(doc), id_field=id_field, dto_model=model)
```
Cito, line 37: is the deepcopy needed here?

services/ifrs/src/ifrs/migration_logic/_utils.py
```diff
@@ -0,0 +1,210 @@
...
+
+        ```
+        class MyMigration(MigrationDefinition, Reversible):
+            ...
+        ```
+        """
+        raise NotImplementedError()
+
+
+class Reversible(ABC):
```
Cito, line 204: Or maybe a ReversibleMigrationDefinition as an ABC subclass of MigrationDefinition with the additional unapply.

But using a mixin is good, too.

services/ifrs/src/ifrs/migration_logic/_manager.py
TheByronHimes: Yes, I like that

services/ifrs/src/ifrs/migration_logic/_manager.py
TheByronHimes: I can try it out. Something like that did come to mind, but I haven't done it (only seen it in the auth-service, for example)

services/ifrs/src/ifrs/migration_logic/_utils.py
TheByronHimes: When I used a regular copy it still updated `doc` and resulted in failing validation since `doc` would then have the id field on it

services/ifrs/src/ifrs/migration_logic/_utils.py
TheByronHimes: Could also do that, yes. I also looked at an array of more involved options to get the exact behavior I envisioned, but in the end it was becoming an overengineered time waste. This simple solution is good I think

services/ifrs/src/ifrs/migration_logic/_manager.py
```diff
@@ -0,0 +1,418 @@
...
+
+        Logs and raises any error that occurs while updating the lock document.
+        """
+        if self._lock_acquired:
+            log.debug("Database lock already acquired")
+            return
+
+        try:
+            lock_col = self.db[self.config.lock_collection]
+            lock_acquired = await lock_col.find_one_and_update(
```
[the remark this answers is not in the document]
TheByronHimes: Currently refactoring the setup logic, so I'll think about that.

services/ifrs/src/ifrs/migration_logic/_utils.py
Cito: It depends on whether we want to encourage devs to write reversible migrations or not. If yes, then the class should always have both and devs should raise NotImplemented if they do not want to implement unapply. 

But I think in practice it will be sufficient to only implement the forward migration. In that case, the current approach is better.

TheByronHimes: The NotImplemented path could lead to the situation where we revert a version or two only to hit the error once some changes have been committed. This lets us check that the method is implemented and is presumably complete before running anything.

services/ifrs/src/ifrs/migration_logic/_manager.py
```diff
@@ -0,0 +1,418 @@
...
+        if not self._lock_acquired:
+            log.debug("Database lock already released")
+            return
+        try:
+            lock_col = self.db[self.config.lock_collection]
+            await lock_col.find_one_and_update(
+                {"lock_acquired": True},
+                {"$set": {"lock_acquired": False, "acquired_at": ""}},
+            )
+            self._lock_acquired = False
```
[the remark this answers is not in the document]
TheByronHimes: Refactored out of existence

services/ifrs/src/ifrs/migration_logic/_manager.py
TheByronHimes: I turned it into a TypedDict and flattened it. The old thinking was that we might change our minds a few times about what info to track for migrations, and it could all get placed into `details` without any side effects.

services/ifrs/src/ifrs/migration_logic/_manager.py
```diff
@@ -0,0 +1,384 @@
...
+                    db=self.db,
+                    unapplying=unapplying,
+                    is_final_migration=is_final_migration,
+                )
+
+                # Call apply/unapply based on migration type
+                await migration.unapply() if unapplying else await migration.apply()
+        except BaseException as exc:
+            error = MigrationStepError(
+                current_ver=current_ver, target_ver=self.target_ver, err_info=str(exc)
```
mephenor, line 326: `current_ver` here is always the starting version passed in from outside.
Wouldn't it be more informative to also log the version from the loop to see where things have failed?

services/ifrs/src/ifrs/migration_logic/_manager.py
```diff
@@ -0,0 +1,384 @@
...
+        Raises `MigrationError` if unsuccessful.
+        """
+        seq = self._get_sequence(current_ver=current_ver, target_ver=self.target_ver)
+        migrations = [self._fetch_migration_cls(ver) for ver in seq]
+        unapplying = self._migration_type == "BACKWARD"
+        try:
+            # Execute & time each migration in order to get to the target DB version
+            for v, migration_cls in zip(seq, migrations, strict=True):
+                # Determine if this is the last migration to apply/unapply
+                last_ver_called = self.target_ver + 1 if unapplying else self.target_ver
```
mephenor, line 312: Based on `_get_sequence` this does not seem right.
Shouldn't this be `self.target_ver - 1` for the backward, unapplying case and `self.target_ver` for the forward case?

services/ifrs/src/ifrs/migration_logic/_manager.py
```diff
@@ -0,0 +1,384 @@
...
+
+            start = time()
+            await self._perform_migrations(current_ver=version)
+            duration_ms = duration_in_ms(time() - start)
+
+            # record the db version
+            await self._record_migration(
+                version=self.target_ver,
+                total_duration_ms=duration_ms,
+            )
```
mephenor, line 374: This means there is only one migration record, even if multiple versions are involved and only the target version is recorded?

services/ifrs/src/ifrs/migration_logic/_manager.py
```diff
@@ -0,0 +1,384 @@
...
+        return True
+
+    async def migrate_or_wait(self):
+        """Try to migrate the database or wait until migrations are completed."""
+        if not self._entered:
+            raise RuntimeError("MigrationManager must be used as a context manager")
+
+        # need to implement some kind of total time limit, warning logging, etc. later
+        while not await self._migrate_db():
+            sleep(self.config.migration_wait_sec)
```
mephenor, line 384: This should probably be `asyncio.sleep` instead

services/ifrs/src/ifrs/migration_logic/_utils.py
```diff
@@ -0,0 +1,210 @@
...
+
+        temp_new_coll_name = self.new_temp_name(coll_name)
+        temp_new_collection = self._db[temp_new_coll_name]
+
+        # naive implementation - update to use batching and bulk inserts
+        async for doc in old_collection.find():
+            output_doc = await method(doc)
+
+            # do validation against model only if we're on the last migration because
+            # the model defined in code may is not guaranteed to match until that time
```
mephenor, line 127:
```suggestion
            # the model defined in code is not guaranteed to match until that time
```

services/ifrs/src/ifrs/migration_logic/_utils.py
```diff
@@ -0,0 +1,210 @@
...
+            if validation_model and (self._is_final_migration or force_validate):
```
mephenor, line 128: This should also check if there even is a model to validate against

services/ifrs/src/ifrs/migration_logic/_utils.py
```diff
@@ -0,0 +1,210 @@
...
+        """Rename old collections to temporarily move them aside without dropping them.
+        Remove temporary prefix from updated collections.
+
+        Assumes apply or unapply has completed.
+        """
+        if isinstance(coll_names, str):
+            coll_names = [coll_names]
+        for coll_name in coll_names:
+            await self.stage_collection(coll_name)
+        self._collections_staged = True
```
mephenor, line 156: What's the impact when encountering an exception before this is set to `True`?
As this isn't persisted, `change_function` could possibly modify already changed collections on the next run, so shouldn't `stage_collection` check for the existence of the old db and do nothing in that case?

Not sure I get the whole logic here and if there's something else that takes care of that in the manager or somewhere else.

services/ifrs/src/ifrs/migration_logic/ifrs_migrations.py
```diff
@@ -0,0 +1,72 @@
...
+        )
+        await self.stage_new_collections(METADATA_COLLECTION)
+        await self.drop_old_collections(METADATA_COLLECTION, enforce_indexes=False)
+
+    async def unapply(self):
+        """Remove `object_size`"""
+        await self.migrate_docs_in_collection(
+            coll_name=METADATA_COLLECTION,
+            change_function=self.remove_object_size,
+            validation_model=FileMetadata,
```
mephenor, line 68: This will only work if the model is changed back to its previous state before running the backward migration.

`validate_doc` should add a hint to check the `validation_model` in its error message.

services/ifrs/tests_ifrs/test_migrations.py
```diff
@@ -0,0 +1,120 @@
...
+        sizes.append(size)
+        coll.insert_one(metadata_base)
+
+        # Upload a temp file object with the given size to our S3 fixture
+        with temp_file_object(
+            bucket_id="permanent", object_id=f"objectid00{n}", size=size
+        ) as f:
+            await s3.populate_file_objects([f])
+
+    post_insert = [_ for _ in coll.find()]
```
mephenor, line 80:
```suggestion
    post_insert = [*coll.find()]
```

services/ifrs/tests_ifrs/test_migrations.py
```diff
@@ -0,0 +1,120 @@
...
+    assert len(post_insert) == 9
+    for doc in post_insert:
+        assert "object_size" not in doc
+
+    # Run the database migrations, which should create the lock/versioning collection
+    #  and then update the 'file_metadata' collection.
+    await run_db_migrations(config=config)
+
+    # Make sure the docs from file_metadata were updated with the right size values
+    post_apply = [_ for _ in coll.find()]
```
mephenor, line 90:
```suggestion
    post_apply = [*coll.find()]
```

services/ifrs/tests_ifrs/test_migrations.py
```diff
@@ -0,0 +1,120 @@
...
+        assert "object_size" in doc
+        assert doc["object_size"] == size
+
+    # Unapply the v2 migration, which will fail because the current model requires
+    #  the `object_size` field.
+    with pytest.raises(MigrationStepError):
+        await run_db_migrations(config=config, target_version=1)
+
+    # Make sure the docs from file_metadata were updated with the right size values
+    post_failed_unapply = [_ for _ in coll.find()]
```
mephenor, line 102:
```suggestion
    post_failed_unapply = [*coll.find()]
```

services/ifrs/tests_ifrs/test_migrations.py
```diff
@@ -0,0 +1,120 @@
...
+    # Monkeypatch the pydantic model to resemble the old version so we can test reversal
+    class PatchedFileMetadata(FileMetadataBase):
+        object_id: str
+
+    monkeypatch.setattr(
+        "ifrs.migration_logic.ifrs_migrations.FileMetadata", PatchedFileMetadata
+    )
+    await run_db_migrations(config=config, target_version=1)
+
+    post_unapply = [_ for _ in coll.find()]
```
mephenor, line 117:
```suggestion
    post_unapply = [*coll.find()]
```

services/ifrs/tests_ifrs/test_migrations.py
```diff
@@ -0,0 +1,120 @@
...
+        assert "object_size" not in doc
+
+    # Run the database migrations, which should create the lock/versioning collection
+    #  and then update the 'file_metadata' collection.
+    await run_db_migrations(config=config)
+
+    # Make sure the docs from file_metadata were updated with the right size values
+    post_apply = [_ for _ in coll.find()]
+    assert len(post_apply) == 9
+    for doc, size in zip(post_apply, sizes, strict=True):
```
mephenor, line 92: Is document order guaranteed here?

mephenor (changes requested):

services/ifrs/src/ifrs/migration_logic/_manager.py
```diff
@@ -0,0 +1,418 @@
...
+        examples=["ifrsLock", "dcsLock"],
+    )
+    db_version_collection: str = Field(
+        ...,
+        description="The name of the collection containing DB version information for this service",
+        examples=["ifrsDbVersions"],
+    )
+    migration_wait_sec: int = Field(
+        ...,
+        description="The number of seconds to wait before checking the DB version again",
```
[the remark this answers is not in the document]
TheByronHimes: No, this is to wait for in-progress migration work to finish before checking the database version again. 
See DB needs migration -> fail to acquire lock -> check back after every `migration_wait_sec` until DB is updated -> start service

services/ifrs/src/ifrs/migration_logic/_manager.py
```diff
@@ -0,0 +1,384 @@
...
+        try:
+            version_coll = self.db[coll_name]
+            with suppress(DuplicateKeyError):
+                await version_coll.insert_one(
+                    {
+                        "_id": 0,
+                        "lock_acquired": True,
+                        "acquired_at": now_as_utc().isoformat(),
+                    }
+                )
```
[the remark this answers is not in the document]
TheByronHimes: If you're referring to the `suppress` context manager, yes. If it's a DuplicateKeyError, it will exit without setting self._lock_acquired to True but otherwise continue. Other error types will be caught by the surrounding try/except and be re-raised as a DbLockError

services/ifrs/src/ifrs/migration_logic/ifrs_migrations.py
TheByronHimes: You're right, which is the reason for only performing validation if the model is supplied _and_ it's the last migration. If a reversion is the result of new commits, then the error serves to inform us that we forgot to change the model back. If the reversion is the result of deploying an older docker image or something functionally equivalent, then an error will not be due to the model (as it will already be what it should be for that older version). I will add the hint -- it's a good idea.

services/ifrs/src/ifrs/migration_logic/_manager.py
TheByronHimes: I verified the behavior is correct locally but decided to simplify it to avoid future confusion. See [606d0bd](https://github.com/ghga-de/file-services-backend/pull/85/commits/606d0bdf2b4c0fdc00ccfa06bcef0baf1a61e8ff)

services/ifrs/src/ifrs/migration_logic/_manager.py
TheByronHimes: Good idea. I moved the `try/except` inside the loop to do this (rather than just adding a log). See [ef84aed](https://github.com/ghga-de/file-services-backend/pull/85/commits/ef84aedc133230dceeeeda3e6a25e44107e19aa5)

services/ifrs/src/ifrs/migration_logic/_manager.py
TheByronHimes: Done

services/ifrs/src/ifrs/migration_logic/_manager.py
```diff
@@ -0,0 +1,418 @@
...
+            return False
+
+        # Initialize db version collection
+        await self._record_migration(
+            version=1,
+            total_duration_ms=duration_in_ms(time() - init_start),
+        )
+        return True
+
+    def _get_sequence(self, *, current_ver: int, target_ver: int) -> list[int]:
```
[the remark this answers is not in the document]
TheByronHimes: You're right. Fixed in [a574067](https://github.com/ghga-de/file-services-backend/pull/85/commits/a574067b8d8900730753403cab4c958b1bf38540)

services/ifrs/src/ifrs/migration_logic/_manager.py
```diff
@@ -0,0 +1,418 @@
...
+        """Migrate forward or backward to reach target DB version.
+
+        Raises `MigrationError` if unsuccessful.
+        """
+        seq = self._get_sequence(current_ver=current_ver, target_ver=self.target_ver)
+        migrations = [self._fetch_migration_cls(ver) for ver in seq]
+        unapplying = self._migration_type == "BACKWARD"
+        try:
+            # Execute & time each migration in order to get to the target DB version
+            for v, migration_cls in zip(seq, migrations, strict=True):
```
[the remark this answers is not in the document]
TheByronHimes: Updated in [a574067](https://github.com/ghga-de/file-services-backend/pull/85/commits/a574067b8d8900730753403cab4c958b1bf38540)

services/ifrs/tests_ifrs/test_migrations.py
TheByronHimes: Good point, no. I changed it so that it stores the generated test sizes in a dict and looks them up by id: [9a4198d](https://github.com/ghga-de/file-services-backend/pull/85/commits/9a4198df933a1d0f60f4518534cd9740b28f7499)

services/ifrs/src/ifrs/migration_logic/_utils.py
TheByronHimes: You're right -- this needs better error checking!

services/ifrs/src/ifrs/migration_logic/ifrs_migrations.py
TheByronHimes: See [8eac31e](https://github.com/ghga-de/file-services-backend/pull/85/commits/8eac31ec267be198dd3012fcf26f0cad45ddbd21)

services/ifrs/src/ifrs/migration_logic/_utils.py
```diff
@@ -63,15 +65,61 @@ def __init__(
...
+
+        # Tracks which collections have had indexes copied over from old collections
+        self._indexes_copied: set[str] = set()
+
+        # Tracks which collections have been staged
+        self._staged_collections: set[str] = set()
+
+    @asynccontextmanager
+    async def auto_finalize(self, coll_names: str | list[str], copy_indexes: bool):
+        """Use within `apply()` or `unapply()` to
```
mephenor, line 84: There's something missing here

services/ifrs/src/ifrs/migration_logic/_utils.py
```diff
@@ -63,15 +65,61 @@ def __init__(
...
+        recommendation to restore the database.
+        """
+        try:
+            # Yield to run the actual migration
+            yield
+            await self.stage_new_collections(coll_names)
+
+            # copy indexes if needed (not implemented yet)
+            if copy_indexes:
+                pass
```
mephenor, line 100: With this not implemented, wouldn't it make more sense to make not copying the indexes the default case?
Silently not doing anything, not even logging, doesn't match the expectation.

services/ifrs/src/ifrs/migration_logic/_utils.py
```diff
@@ -110,78 +158,106 @@ async def migrate_docs_in_collection(
...
-        if self._collections_staged:
+        if coll_name in self._staged_collections:
             raise RuntimeError("Collections already staged, changes shouldn't be made.")
 
         old_collection = self._db[coll_name]
         method = change_function
 
+        # Drop the temp collection first to make sure we're starting fresh.
         temp_new_coll_name = self.new_temp_name(coll_name)
+        await self._db.drop_collection(temp_new_coll_name)
```
mephenor, line 169: Is this a no-op if the collection does not exist?

services/ifrs/src/ifrs/migration_logic/_utils.py
```diff
@@ -63,15 +65,61 @@ def __init__(
...
+        a more manual approach. For that reason, this context manager is optional.
+
+        If an error occurs during the migration process, staged changes will be unstaged
+        and dropped. If a subsequent error occurs during cleanup, it is logged with a
+        recommendation to restore the database.
+        """
+        try:
+            # Yield to run the actual migration
+            yield
+            await self.stage_new_collections(coll_names)
```
mephenor, line 96: As this is already working outside of hexkit, wouldn't it be easier to wrap it in a transaction so we have no partially staged/unstaged collections and MongoDB deals with the rollback?

services/ifrs/src/ifrs/migration_logic/_utils.py
TheByronHimes: 😵 yep

services/ifrs/src/ifrs/migration_logic/_utils.py
```diff
@@ -0,0 +1,286 @@
...
+        # e.g. "users" -> "tmp_v7_old_users"
+        old_collection = self._db[original_coll_name]
+        temp_old_coll_name = self.old_temp_name(original_coll_name)
+        await old_collection.rename(temp_old_coll_name)
+
+        # Rename the new, temp collection by removing its prefix
+        # e.g. "tmp_v7_new_users" -> "users"
+        temp_new_coll_name = self.new_temp_name(original_coll_name)
+        new_collection = self._db[temp_new_coll_name]
+        await new_collection.rename(original_coll_name)
```
mephenor, line 201: Might be easier to parse mentally if both have the same order of actions.

services/ifrs/src/ifrs/migration_logic/_utils.py
```diff
@@ -0,0 +1,286 @@
...
+            True for modifications that don't involve changes to the collections' indexes.
+        """
+        if enforce_indexes and not self._indexes_copied:
+            raise RuntimeError("Indexes have not been applied to staged collections")
+
+        for coll_to_drop in list(self._staged_collections):
+            old_temp_name = self.old_temp_name(coll_to_drop)
+            collection = self._db[old_temp_name]
+            await collection.drop()
+            log.debug("Dropped old collection")
```
mephenor, line 259: Log statement should print the name of the dropped collection.

services/ifrs/src/ifrs/migration_logic/_utils.py
TheByronHimes: Yes

services/ifrs/src/ifrs/migration_logic/_utils.py
TheByronHimes: Transactions like this can only be done on replica sets, which A) I was not able to get to work with the testcontainer setup we have (for now) and B) would require changes to hexkit to make the testing valid. I agree though that it would be the right solution if it were more doable. We should revisit it in the future as an improvement/refactoring item

services/ifrs/src/ifrs/migration_logic/_utils.py
TheByronHimes: Fixed in [2c1644c](https://github.com/ghga-de/file-services-backend/pull/85/commits/2c1644c68ac9f38d1101a20190cb8fa5d3fa104f)

services/ifrs/src/ifrs/migration_logic/_utils.py
TheByronHimes: See [2c1644c](https://github.com/ghga-de/file-services-backend/pull/85/commits/2c1644c68ac9f38d1101a20190cb8fa5d3fa104f). Made it log the original name as well as the temp table name, e.g. "users" and "temp_v2_old_users".

services/ifrs/src/ifrs/migration_logic/_utils.py
TheByronHimes: I've replaced `pass` with a `NotImplementedError` and made the parameter have a default of False.

services/ifrs/src/ifrs/migration_logic/_utils.py
```diff
@@ -97,7 +102,7 @@ async def auto_finalize(self, coll_names: str | list[str], copy_indexes: bool):
 
             # copy indexes if needed (not implemented yet)
             if copy_indexes:
-                pass
+                raise NotImplementedError("Index copying is not yet implemented")
```
mephenor, line 105: One last thing here:
If this raises, it would be better to move this check before the yield, so this fails early.
Else this always tries to do the migration first only to unstage all changes again.

mephenor (changes requested):

services/ifrs/src/ifrs/migration_logic/_utils.py
TheByronHimes: fair enough. Done

services/ifrs/src/ifrs/migration_logic/_manager.py
TheByronHimes: Will leave this for the actual implementation if we do it

mephenor (approved):

PR #97, opened 2025-02-26
UCS: Enable DLQ - (GSI-1398)

TheByronHimes:
The UCS version was not bumped because I want to include it in the next monorepo release.

To keep everything clear, this PR is only concerned with the **consumer** side of Kafka. I.e. this doesn't touch outbox publishing.

Changes include:
- replaced all usage of the `KafkaOutboxSubscriber` in the UCS with the `KafkaEventSubscriber `by bundling the existing outbox subscriber _translators_ with the event subscriber translator via the new `ComboTranslator` from `hexkit 4.1.0`.
  - except for one test in `test_outbox_dao.py` where it's easier to keep it like it is
- Added a test module with two tests for the DLQ functionality
- Updated the `inject.py` module
- removed the outbox_subscriber from the joint fixture
- Updated the testing `get_config()` function to take arbitrary max-priority kwarg overrides to get around the default kafka fixture config (I've done this in all the other DLQ PRs so far, too)


services/ucs/src/ucs/main.py
```diff
@@ -46,12 +45,8 @@ async def consume_events(run_forever: bool = True):
 
     async with (
         prepare_event_subscriber(config=config) as event_subscriber,
-        prepare_outbox_subscriber(config=config) as outbox_subscriber,
     ):
-        await asyncio.gather(
-            event_subscriber.run(forever=run_forever),
-            outbox_subscriber.run(forever=run_forever),
-        )
+        await asyncio.gather(event_subscriber.run(forever=run_forever))
```
mephenor:
```suggestion
        await event_subscriber.run(forever=run_forever)
```
Gather is no longer needed as this is back to awaiting one subscriber

mephenor (changes requested):

services/ucs/src/ucs/main.py
TheByronHimes: That should have already been removed in [d47b2e9](https://github.com/ghga-de/file-services-backend/pull/97/commits/d47b2e91040ff35535dc0428977071fc34a42a54)

mephenor (approved):

PR #105, opened 2025-04-01
Opentelemetry instrumentation

mephenor:
This PR adds OpenTelemetry support and some miscellaneous updates from the microservice template repo.
OpenTelemetry instrumentation is achieved by using the corresponding instrumentation libraries and autoinstrumenting the services via `opentelemetry-instrument`.
While there exists a way to initiate autoinstrumentation programmatically since the latest OpenTelemetry release, it is either not fully equivalent or sensitive to import order, not covering all the intra- and interservice spans that were visible during eploration.

Most of the functions at the service boundaries, i.e. those associated with FastAPI routes, outgoing httpx calls, event subscription or publication, open their own spans.
This is not strictly necessary, but might improve readability by providing a mapping between spans and entry/exit points in the services.


Cito (approved):
Looks good.

I made some suggestions to make OT support a more DRY and easier to maintain. But you can already merge as is and tackle this later.

PR #108, opened 2025-04-16
All services/OpenTelemetry refactoring

mephenor:
This PR includes several changes, some not fully related to OpenTelemetry:
 - OpenTelemetry related functionality now uses the hexkit helpers
 - Added a section to the readme describing some of the environment variables for OpenTelemetry
 - ~~Service names have been brought in line across all services + now reside in `constants.py` to avoid circular import issues~~
 - Updated test workflow to re-enable coverage statistics. This will probably run into some issues when only a subset of tests is run and will need to be revisited once coveralls has their `carry forward` feature released on stable. For now this will show coverage based on the tests actually run. See https://coveralls.io/github/ghga-de/file-services-backend?branch=all_services/opentelemetry_refactoring for more details on how that actually looks like on this branch.
 
~~No version bumps included, waiting to see what happens with the current release before attaching a version number to these changes.~~ Bumped minor versions for all services due to new config options.


TheByronHimes:
Moving the service name default to a constants file is fine (maybe slightly misleading), but importing that constant and using it in the spans means that the span value and configured value will be out of sync if the configured service name differs.

services/pcs/src/pcs/inject.py
```diff
@@ -54,11 +50,7 @@ async def get_persistent_publisher(
 @asynccontextmanager
 async def prepare_core(*, config: Config) -> AsyncGenerator[FileDeletionPort, None]:
     """Construct and initialize the core component and its outbound dependencies."""
-    resource = Resource(attributes={SERVICE_NAME: "Purge Controller Service"})
-    trace_provider = TracerProvider(resource=resource)
-    processor = BatchSpanProcessor(OTLPSpanExporter())
-    trace_provider.add_span_processor(processor)
-    trace.set_tracer_provider(trace_provider)
+    configure_tracer(service_name="Purge Controller Service")
```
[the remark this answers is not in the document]
mephenor: Not having that in the entrypoint might have been a side-effect of testing the programmatic autoinstrumentation previously. Need to have another look.

About the naming: The other services also had the full name here, but this can be changed to source from config instead.

Setting the service name from the actual config should be possible the SpanTracers to singletons.

services/dcs/src/dcs/core/data_repository.py
TheByronHimes, line 1: I would have expected to see spans started on `access_drs_object()`, `cleanup_outbox`, etc. What are the criteria for determining which functions get the decorator or don't?

services/ekss/src/ekss/adapters/outbound/vault/client.py
TheByronHimes, line 1: Same question as in the DCS, since here some methods in the `VaultAdapter` are tagged and others aren't, and it doesn't match my intuition, which is "any function that does something worth tracing".

TheByronHimes (commented):
Need more info and then I can do a better review:

services/dcs/src/dcs/core/data_repository.py
mephenor: Currently the main concern was adding additional spans at the service boundaries so 1) event subscribers attach their newly opened span to a parent span coming from another service, i.e. the trace doesn't stop at the boundary on the receiving side, 2) spans are named in a way that makes tracing them back to the actual function easy. Some instrumentation libaries provide rather generic names for the spans they create, e.g. something like `event sub`.

Within the core logic we get most of the benefit right now from autoinstrumentation for mongodb and boto3 which open new spans for their respective interactions. If we need additional insight in between to better delineate where something might be going wrong, we can always add that later. 
I'm not sure about how much of this is actually currently enabled in production, so if even if this is a bit conservative right now, we should have some feedback on where more spans might be beneficial soon enough.

The goal of this PR is only to refactor the messy initial version of the implementation and there's also some experimentation ongoing, i.e. changing the default to not attach exceptions to a span and not set the error state by default. 
The second thing is there to see if this will reset the state of e.g. a boto head object call to check if an object exists. The instrumentation sets the trace to errored if the object is not found, which is not always an error in our use case.
The `start_span` decorator exposes two arguments that allow to gradually opt into the previous default behaviour once we're sure how that works and that the attached exceptions don't expose sensitive informations.

TheByronHimes (approved):

PR #120, opened 2025-07-28
DCS Hexkit v6 + Migrations (GSI-1780)

TheByronHimes:
Ignore the failed mypy check because it's only flagging errors in other services. This will occur for all the PRs until they are merged into one.

### DCS Changes:
- V2 migration that covers two collections: `drs_objects` and `dcsPersistedEvents`.
- Models updated to use UUID4 and UTCDatetime where appropriate
- Use of `now_as_utc` replaced with less precise `now_utc_ms_prec`
- `MongoDbDaoFactory` is now used as a context manager

### DIFF
--- a/services/dcs/src/dcs/migrations/definitions.py
+++ b/services/dcs/src/dcs/migrations/definitions.py
@@ -1,0 +1,74 @@
+# Copyright 2021 - 2025 Universität Tübingen, DKFZ, EMBL, and Universität zu Köln
+# for the German Human Genome-Phenome Archive (GHGA)
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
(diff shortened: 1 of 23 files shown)


services/dcs/src/dcs/migrations/definitions.py
```diff
@@ -0,0 +1,129 @@
...
+                doc[field] = doc[field].isoformat()
+            return doc
+
+        async def revert_persistent_event(doc: Document) -> Document:
+            """Convert the fields back into strings"""
+            doc.pop("event_id")
+            doc["correlation_id"] = str(doc["correlation_id"])
+            doc["created"] = doc["created"].isoformat()
+
+            if payload := doc["payload"]:
```
mephenor, line 110: See above

services/dcs/src/dcs/migrations/definitions.py
```diff
@@ -0,0 +1,129 @@
...
+    - drs_objects (AccessTimeDrsObject)
+        - object_id, creation_date, and last_accessed
+    - dcsPersistedEvents
+
+    This can be reversed by converting the UUIDs and datetimes back to strings.
+    """
+
+    version = 2
+
+    _object_id: str = "object_id"
```
mephenor, line 50: Bit of an inconsistency to have that defined as class variable, but not `upload_date`, etc.
Would use the string directly to be consistent with the rest of the code.

mephenor (changes requested):

services/dcs/src/dcs/migrations/definitions.py
TheByronHimes: Guilty. It is removed

mephenor (approved):

PR #129, opened 2025-08-21
FIS: Fix v3 migration for funky correlation IDs

TheByronHimes:
There were some non-v4 UUID string values for the correlation ID in the FIS persisted events collection. We think these might have come from the api gateway in before we switched. The values are all older and cutoff after a certain date. Therefore, the solution for now is to simply replace these values in the migration when encountered, log the switch, and keep an eye out for the issue in the future. If it comes up again, we will perform a more thorough investigation.


dontseyit:
Thank you quick fix

PR #136, opened 2025-09-30
[DCS] Adjust WorkOrderTokenContext/Config type field name to match WPS changes

mephenor:


dontseyit (approved):
Thank you for the quick fix

PR #152, opened 2026-01-26
FIS: Overhaul for Upload Path Pt. 2 (GSI-1992)

TheByronHimes:
- Migration tests were removed because we're dropping the old FIS data. All files are registered in IFRS anyway, so the FIS's database contents are not needed.
- `run_db_migrations()` was moved into a standalone entrypoint command so we can use it in an `init` container.
- For clarity, `data_hub` is no longer a thing. Just `storage_alias`.

Since the (re?)introduction of `object_id`, the FileUnderInterrogation collection now has an index over `object_id`. This is because when DHFS asks if files can be deleted from the interrogation bucket, it now does so using what corresponds to the object ID.


mephenor (commented):
Skipped checking the test code a

services/fis/src/fis/core/interrogation.py
```diff
@@ -0,0 +1,255 @@
...
+        elif file.state == "inbox":
+            with suppress(ResourceAlreadyExistsError):
+                await self._dao.insert(file)
+                return
+
+        # If we make it past that block, then it is not new and we must compare.
+        local_file = await self._dao.get_by_id(file.id)
+
+        # Ignore if outdated
+        if local_file.state_updated >= file.state_updated:
```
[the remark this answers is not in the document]
TheByronHimes: Yeah, `state_updated` is a timestamp of when `state` was set.

mephenor (approved):

services/fis/src/fis/adapters/outbound/dao.py
```diff
@@ -15,15 +15,17 @@
...
 from fis.ports.outbound.dao import FileDao
 
 
 async def get_file_dao(*, dao_factory: DaoFactoryProtocol) -> FileDao:
     """Setup the DAOs using the specified provider of the DaoFactoryProtocol."""
     return await dao_factory.get_dao(
-        name="ingestedFiles",
-        dto_model=FileIdModel,
-        id_field="file_id",
+        name="interrogatedFiles",
```
mephenor: Slightly confused about the collection name here, as `FileUnderInterrogation` has `interrogated` set to `False` by default.

services/fis/tests_fis/fixtures/joint.py
```diff
@@ -15,117 +15,102 @@
...
+
+
+@pytest.fixture()
+def rig(config: Config):
+    """Produce a populated JointRig instance"""
+    config = config
+    event_store = InMemEventStore()
+    interrogation_handler = InterrogationHandler(
+        config=config,
+        file_dao=(dao := InMemFileDao()),
```
mephenor: Not a fan of the walrus in this case

mephenor (changes requested):

services/fis/src/fis/core/interrogation.py
```diff
@@ -0,0 +1,264 @@
...
+            return True
+        except MultipleHitsFoundError as err:
+            # This should never happen, and if it does then we need eyes on it.
+            msg = f"Found multiple files with the same object_id of {object_id}"
+            log.critical(msg, extra={"object_id": object_id})
+            raise RuntimeError(msg) from err
+        else:
+            return file.can_remove
+
+    async def handle_interrogation_report(self, *, report: models.InterrogationReport):
```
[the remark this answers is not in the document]
TheByronHimes: @mephenor Yes, thanks for seeing this. We can handle duplicate submissions with a simple check like you suggest. The latter concern could be mitigated by making changes in both FIS and DHFS:
- FIS: Do basic idempotence check to guard against duplicate report submissions. Additionally, raise an error if a report is received with a different secret ID so we can get eyes on it.
- DHFS: Instead of deleting re-encrypted files from the interrogation bucket when report submission fails, don't do any status code interpretation or substantial error handling (just basic catch/log/reraise). Leave the files there in the bucket - they'll get deleted if/when the same file is interrogated anew. Either FIS gets the report and updates its DB accordingly or it doesn't. The result of that POST call should ultimately flow back to the DHFS via subsequent calls to `.get_removable_files()` and `.fetch_new_uploads()`. This prevents the different secret ID problem, which is the most problematic inconsistency mode.

For now we should assume (and effectuate via policy/comms) that the DHFS runs as a single instance. Preventing race conditions in order to allow parallel instances at a given Data Hub should be an upgrade in the middle future.

services/fis/src/fis/adapters/outbound/secrets.py
mephenor: Maybe something for another service-commons ticket, but with that repetition seen here and in other places using the client, this could probably use some kind of more generic error handling in between that unifies error sources and exposes them in one type.

services/fis/src/fis/core/interrogation.py
```diff
@@ -83,68 +79,170 @@ async def check_if_removable(self, *, object_id: UUID4) -> bool:
...
+                    file_id,
+                    extra={"file_id": file_id, "differing_fields": different_fields},
                 )
-            except tenacity.RetryError as err:
-                exception = err.last_attempt.exception()
-                reason = (
-                    str(exception.args[0])
-                    if exception and exception.args
-                    else "Unknown"
+                # Don't `raise...from err` because the above did not actually cause this
```
mephenor: Comment can be removed, this is in the else clause, thus there's no error context to raise from

mephenor (changes requested):

services/fis/src/fis/core/interrogation.py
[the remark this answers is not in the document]
TheByronHimes: The pattern has been reworked for simplicity, but if the service spontaneously dies at certain points, there's no way around it but to restore from a backup or manually fix the db.

mephenor (approved):

PR #155, opened 2026-02-24
DCS: Upload Path Pt. 2 (GSI-2154)

TheByronHimes:
Migrations:
- There is no v3 migration, and the tests for v2 were removed. Left the definition for posterity, for now. We need to drop all the DCS collections except for the DB versioning collection

Changes include:
- Registered objects are keyed by the UUID4 file_id field now
- Accession is absent from DCS _except_ for when it is supplied by the user/connector - the DCS HTTP endpoint path parameters still expect the accession number, but internally everything is done based on file_id. 
- The WOT supplied by the WPS by way of the connector _used_ to have `file_id: str = <accession>`, but now it contains two fields: `accession: str` & `file_id: UUID4`. 
- Some parameters and variables were renamed across the service but retain their original meaning and/or function.
- The EKSS logic has been moved into a dedicated port/adapter (`SecretsClient`) similarly to how things are now done in other recently updated file services.


services/dcs/src/dcs/adapters/outbound/event_pub.py
```diff
@@ -93,34 +93,33 @@ async def download_served(
...
 
     @TRACER.start_as_current_span("EventPubTranslator.file_registered")
     async def file_registered(self, *, drs_object: models.DrsObjectWithUri) -> None:
         """Communicates the event that a file has been registered."""
-        payload = event_schemas.FileRegisteredForDownload(
+        payload = models.FileRegisteredForDownload(
             file_id=drs_object.file_id,
             decrypted_sha256=drs_object.decrypted_sha256,
-            upload_date=drs_object.creation_date,
-            drs_uri=drs_object.self_uri,
```
mephenor, line 106: What was that variable again?

services/dcs/src/dcs/core/data_repository.py
```diff
@@ -127,24 +112,26 @@ def __init__(
...
 
     def _get_model_with_self_uri(
-        self, *, drs_object: models.DrsObject
+        self, *, drs_object: models.DrsObject, accession: str
     ) -> models.DrsObjectWithUri:
         """Add the DRS self URI to an DRS object."""
         return models.DrsObjectWithUri(
             **drs_object.model_dump(),
-            self_uri=self._get_drs_uri(drs_id=drs_object.file_id),
+            self_uri=self._get_drs_uri(drs_id=accession),
```
mephenor: `access_drs_object` below moved to renaming this to accession, could also do this here to be consistent

services/dcs/src/dcs/core/data_repository.py
```diff
@@ -165,38 +152,41 @@ async def _get_access_model(
...
             stopped = perf_counter() - started
             log.debug("Fetched DRS object model in %.3f seconds.", stopped)
         except ResourceNotFoundError as error:
-            drs_object_not_found = self.DrsObjectNotFoundError(drs_id=drs_id)
-            log.error(drs_object_not_found)
+            drs_object_not_found = self.DrsObjectNotFoundError(file_id=file_id)
+            log.error(drs_object_not_found, extra=log_extra)
             raise drs_object_not_found from error
 
+        # TODO: Add more comments in here
```
mephenor: A wild TODO appears

services/dcs/src/dcs/core/data_repository.py
```diff
@@ -258,6 +258,7 @@ async def cleanup_outbox_buckets(
         for storage_alias in object_storages_config.object_storages:
             await self.cleanup_outbox(storage_alias=storage_alias)
 
+    # TODO: Either stick with outbox here or rename but be consistent
```
mephenor: Can be handled in the other outbox related ticket.
We can call it download bucket for consistency

services/dcs/src/dcs/core/data_repository.py
```diff
@@ -442,11 +441,12 @@ async def delete_file(self, *, file_id: str) -> None:
...
+                drs_object.object_id,
             )
 
         # Remove file from database and send success event
         # Should not fail as we got the DRS object by the same ID
-        await self._drs_object_dao.delete(id_=file_id)
+        await self._drs_object_dao.delete(file_id)
         await self._event_publisher.file_deleted(file_id=file_id)
-        log.info(f"Successfully deleted entries for file with id '{file_id}'.")
+        log.info(f"Successfully deleted entries for file '{file_id}'.")
```
mephenor: Can we add the alias and bucket name as extra information here?
Could help in debugging if something should ever go wrong with deletion

services/dcs/src/dcs/core/models.py
```diff
@@ -111,6 +104,146 @@ class DrsObjectResponseModel(BaseModel):
     access_methods: list[AccessMethod]
     checksums: list[Checksum]
     created_time: str
-    id: str
+    id: str  # this is the accession number provided in the request
```
mephenor, line 107: So this is the DRS `object_id`, which corresponds to the `accession`?

services/dcs/src/dcs/core/models.py
```diff
@@ -111,6 +104,146 @@ class DrsObjectResponseModel(BaseModel):
...
+    backend has been made.
+
+    This local definition will be replaced by the `ghga-event-schemas` definition
+    once implemented there.
+    """
+
+    file_id: UUID4 = Field(..., description="Unique identifier for the file")
+
+
+class FileDeletionSuccess(BaseModel):
```
mephenor: Can use `FileDeletionRequested` as base.

services/dcs/src/dcs/config.py
```diff
@@ -56,7 +58,15 @@ class Config(
...
 ):
     """Config parameters and their defaults."""
 
     service_name: str = SERVICE_NAME
+    srs_base_url: str = Field(
+        default=...,
+        description="Base URL of the Study Repository Service for the V3 Migration.",
+        title="SRS base URL",
+        examples=["http://srs:8080/"],
+    )
```
mephenor: This is used where exactly?

services/dcs/tests_dcs/test_edge_cases.py
```diff
@@ -80,37 +82,42 @@ async def test_access_non_existing(joint_fixture: JointFixture):
     """Checks that requesting access to a non-existing DRS object fails with the
     expected exception.
     """
-    file_id = "my-non-existing-id"
+    accession = "GHGADoesNotExist"
```
mephenor, line 85: Lies ;)

services/dcs/src/dcs/adapters/outbound/http/api_calls.py
```diff
@@ -15,70 +15,38 @@
...
+    CompositeCacheConfig,
+    CompositeTransportFactory,
+    ratelimiting_retry_proxies,
+)
+from pydantic import Field
+
+__all__ = ["HttpClientConfig", "get_configured_httpx_client"]
+
+
+class HttpClientConfig(CompositeCacheConfig):
```
mephenor: This doesn't need caching, right?
So this should be `CompositeConfig` instead.

services/dcs/src/dcs/adapters/outbound/http/secrets.py
```diff
@@ -0,0 +1,149 @@
...
+            RequestFailedError: if an error prevents obtaining a response.
+            BadResponseCodeError: if a response is received but the status code
+                indicates that the request was unsuccessful.
+        """
+        receiver_public_key_base64 = base64.urlsafe_b64encode(
+            base64.b64decode(receiver_public_key)
+        ).decode()
+        api_url = f"{self._api_base}/secrets/{secret_id}/envelopes/{receiver_public_key_base64}"
+        try:
+            response = httpx.get(url=api_url)
```
mephenor:
```suggestion
            response = self._httpx_client.get(url=api_url)
```

mephenor (changes requested):

services/dcs/src/dcs/adapters/outbound/event_pub.py
TheByronHimes: Which one are you referring to?

services/dcs/src/dcs/core/data_repository.py
TheByronHimes: If you're referring to renaming _get_drs_uri's drs_id parameter to `accession`, done.

services/dcs/src/dcs/adapters/outbound/event_pub.py
mephenor: The drs_uri.
What exactly does it contain and is there a consumer that relies on its presence?

services/dcs/src/dcs/config.py
TheByronHimes: Thanks for finding that - it's not used. It was for back when the plan was to submit accession maps from DCS to SRS during migration. But we're not doing that anymore, so this should be deleted.

services/dcs/src/dcs/adapters/outbound/http/api_calls.py
TheByronHimes: That was an accident, glad you saw it.

services/dcs/src/dcs/core/models.py
TheByronHimes: Exactly

mephenor (approved):

services/dcs/src/dcs/adapters/outbound/event_pub.py
TheByronHimes: It points to the DCS download endpoint, I believe (the base url comes from config)

PR #158, opened 2026-03-03
Improve DCS storage cleaner ergonomics

mephenor:
### Main
- Adds a remove_dangling_objects flag (default: False) to the CLI of `cleanup_outbox` part of the data_repository. If dangling outbox objects are detected, i.e. no DB entry, but present in S3, setting this to true allows for removal of those objects
- Consolidated some split error messages into the actual `OutboxCleanupError` class and changed log levels, where appropriate
- Renamed outbox -> download bucket. This also affects the public CLI API
- Split cleanup from data repository functionality and moved it into its own core/port modules

### Misc
- Bumps patch version of DCS, doesn't touch repo version
- Adds missing changes to the coveralls part of the test action
- Removes remaining mention of IRS
- Fixes FIS OpenAPI doc issue triggered by action
- Adapts failing test cases to work again

### DIFF
--- a/services/dcs/src/dcs/cli.py
+++ b/services/dcs/src/dcs/cli.py
@@ -24,7 +24,7 @@ from dcs.main import (
     consume_events,
     migrate_db,
     publish_events,
-    run_download_bucket_cleanup,
+    run_download_bucket_cleaner,
     run_rest_app,
 )
 
@@ -44,10 +44,10 @@ def sync_consume_events(run_forever: bool = True):
 
 
 @cli.command(name="cleanup-download-bucket")
-def sync_run_cleanup(remove_dangling_objects: bool = False):
+def sync_run_cleaner(remove_dangling_objects: bool = False):
     """Run download bucket cleanup"""
     asyncio.run(
-        run_download_bucket_cleanup(remove_dangling_objects=remove_dangling_objects)
+        run_download_bucket_cleaner(remove_dangling_objects=remove_dangling_objects)
     )
 
 
--- a/services/dcs/src/dcs/cli.py
+++ b/services/dcs/src/dcs/cli.py
@@ -24,7 +24,7 @@ from dcs.main import (
     consume_events,
     migrate_db,
     publish_events,
-    run_outbox_cleanup,
+    run_download_bucket_cleanup,
     run_rest_app,
 )
 
@@ -43,10 +43,12 @@ def sync_consume_events(run_forever: bool = True):
     asyncio.run(consume_events(run_forever=run_forever))
 
 
-@cli.command(name="cleanup-outbox")
+@cli.command(name="cleanup-donwload-bucket")
 def sync_run_cleanup(remove_dangling_objects: bool = False):
-    """Run outbox cleanup"""
-    asyncio.run(run_outbox_cleanup(remove_dangling_objects=remove_dangling_objects))
+    """Run download bucket cleanup"""
+    asyncio.run(
+        run_download_bucket_cleanup(remove_dangling_objects=remove_dangling_objects)
+    )
 
 
 @cli.command(name="publish-events")
--- a/services/dcs/src/dcs/cli.py
+++ b/services/dcs/src/dcs/cli.py
@@ -43,7 +43,7 @@ def sync_consume_events(run_forever: bool = True):
     asyncio.run(consume_events(run_forever=run_forever))
 
 
-@cli.command(name="cleanup-donwload-bucket")
+@cli.command(name="cleanup-download-bucket")
 def sync_run_cleanup(remove_dangling_objects: bool = False):
     """Run download bucket cleanup"""
     asyncio.run(
--- a/services/dcs/src/dcs/cli.py
+++ b/services/dcs/src/dcs/cli.py
@@ -44,9 +44,9 @@ def sync_consume_events(run_forever: bool = True):
 
 
 @cli.command(name="cleanup-outbox")
-def sync_run_cleanup():
+def sync_run_cleanup(remove_dangling_objects: bool = False):
     """Run outbox cleanup"""
-    asyncio.run(run_outbox_cleanup())
+    asyncio.run(run_outbox_cleanup(remove_dangling_objects=remove_dangling_objects))
 
 
 @cli.command(name="publish-events")
--- a/services/dcs/src/dcs/core/data_repository.py
+++ b/services/dcs/src/dcs/core/data_repository.py
@@ -278,7 +278,8 @@ class DataRepository(DataRepositoryPort):
         """
         # Run on demand through CLI, so crashing should be ok if the alias is not configured
         log.info(
-            f"Starting download bucket cleanup for storage identified by alias {storage_alias}."
+            "Starting download bucket cleanup for storage identified by alias %s.",
+            storage_alias,
         )
         try:
             bucket_id, object_storage = self._object_storages.for_alias(storage_alias)
--- a/services/dcs/src/dcs/core/data_repository.py
+++ b/services/dcs/src/dcs/core/data_repository.py
@@ -19,14 +19,12 @@ import contextlib
 import logging
 import re
 import uuid
-from datetime import timedelta
 from time import perf_counter
 
 from ghga_service_commons.utils.multinode_storage import (
     S3ObjectStorages,
-    S3ObjectStoragesConfig,
 )
-from hexkit.protocols.dao import NoHitsFoundError, ResourceNotFoundError
+from hexkit.protocols.dao import ResourceNotFoundError
 from hexkit.protocols.objstorage import ObjectStorageProtocol
 from hexkit.utils import now_utc_ms_prec
 from pydantic import UUID4, Field, PositiveInt, field_validator
@@ -35,6 +33,7 @@ from pydantic_settings import BaseSettings
 from dcs.adapters.outbound.http import exceptions
 from dcs.constants import TRACER
 from dcs.core import models
+from dcs.core.errors import StorageAliasNotConfiguredError
 from dcs.ports.inbound.data_repository import DataRepositoryPort
 from dcs.ports.outbound.dao import DrsObjectDaoPort
 from dcs.ports.outbound.event_pub import EventPublisherPort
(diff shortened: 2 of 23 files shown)


services/dcs/src/dcs/core/data_repository.py
```diff
@@ -160,8 +160,8 @@ async def access_drs_object(
     ) -> models.DrsObjectResponseModel:
         """
         Serve the specified DRS object with access information.
-        If it does not exists in the outbox, yet, a RetryAccessLaterError is raised that
-        instructs to retry the call after a specified amount of time.
+        If it does not exists in the downloadbucket, yet, a RetryAccessLaterError
```
TheByronHimes:
```suggestion
        If it does not exists in the download bucket, yet, a RetryAccessLaterError
```

services/dcs/src/dcs/core/data_repository.py
```diff
@@ -253,27 +253,32 @@ async def access_drs_object(
...
+        Check if files present in the download bucket have outlived their allocated time
+        and remove all that do.
+        For each file in the download bucket, its 'last_accessed' field is checked and compared
+        to the current datetime. If the threshold configured in the download_bucket_cache_timeout
+        option is met or exceeded, the corresponding file is removed from the download bucket.
         """
         # Run on demand through CLI, so crashing should be ok if the alias is not configured
         log.info(
-            f"Starting outbox cleanup for storage identified by alias {storage_alias}."
+            f"Starting download bucket cleanup for storage identified by alias {storage_alias}."
```
TheByronHimes:
```suggestion
            "Starting download bucket cleanup for storage identified by alias %s.",
            storage_alias,
```

services/dcs/src/dcs/core/data_repository.py
```diff
@@ -282,16 +287,16 @@ async def cleanup_outbox(self, *, storage_alias: str):
                 alias=storage_alias
             )
             log.critical(storage_alias_not_configured)
-            log.warning(
-                f"Skipping outbox cleanup for storage {storage_alias} as it is not configured."
+            log.info(
+                f"Skipping download bucket cleanup for storage {storage_alias} as it is not configured."
             )
```
TheByronHimes:
```suggestion
            log.info(
                "Skipping download bucket cleanup for storage %s as it is not configured.",
                storage_alias,
            )
```

services/dcs/src/dcs/core/data_repository.py
```diff
@@ -305,18 +310,20 @@ async def cleanup_outbox(self, *, storage_alias: str):
...
+                    continue
 
-            # only remove file if last access is later than outbox_cache_timeout days ago
-            if drs_object.last_accessed <= threshold:
+            # only remove file if last access is later than download bucket_cache_timeout days ago
+            if remove_dangling_objects or drs_object.last_accessed <= threshold:
                 log.info(
-                    f"Deleting object {object_id} from bucket {bucket_id} in storage {storage_alias}."
+                    f"Deleting object {object_id} from download bucket {bucket_id} in storage {storage_alias}."
                 )
```
TheByronHimes:
```suggestion
                log.info(
                    "Deleting object %s from download bucket %s in storage %s.",
                    object_id,
                    bucket_id,
                    storage_alias,
                )
```

services/dcs/src/dcs/core/data_repository.py
```diff
@@ -305,18 +310,20 @@ async def cleanup_outbox(self, *, storage_alias: str):
...
+                        storage_alias=storage_alias,
+                        reason="Object not found in database, skipping.",
+                    )
+                    log.warning(cleanup_error)
+                    continue
 
-            # only remove file if last access is later than outbox_cache_timeout days ago
-            if drs_object.last_accessed <= threshold:
+            # only remove file if last access is later than download bucket_cache_timeout days ago
+            if remove_dangling_objects or drs_object.last_accessed <= threshold:
```
TheByronHimes: If we run this command with `remove_dangling_objects = True` then every object will get deleted, even if it hasn't been in there long enough to cross the threshold. We should only use `remove_dangling_objects` to green-light removal in the NoHitsFoundError case, unless I misunderstand the process

services/dcs/src/dcs/cli.py
```diff
@@ -43,10 +43,12 @@ def sync_consume_events(run_forever: bool = True):
     asyncio.run(consume_events(run_forever=run_forever))
 
 
-@cli.command(name="cleanup-outbox")
-def sync_run_cleanup():
-    """Run outbox cleanup"""
-    asyncio.run(run_outbox_cleanup())
+@cli.command(name="cleanup-donwload-bucket")
```
TheByronHimes:
```suggestion
@cli.command(name="cleanup-download-bucket")
```

services/dcs/src/dcs/inject.py
```diff
@@ -162,20 +162,24 @@ async def prepare_event_subscriber(
...
-async def prepare_outbox_cleaner(
+async def prepare_download_bucket_cleaner(
     *,
     config: Config,
     data_repo_override: DataRepositoryPort | None = None,
-) -> AsyncGenerator[OutboxCleaner]:
-    """Construct and initialize a coroutine that cleans the outbox once invoked.
+    remove_dangling_objects: bool = False,
+) -> AsyncGenerator[DownloadBucketCleaner]:
+    """Construct and initialize a coroutine that cleans the downlaod bucket once invoked.
```
TheByronHimes:
```suggestion
    """Construct and initialize a coroutine that cleans the download bucket once invoked.
```

services/dcs/src/dcs/inject.py
```diff
@@ -162,20 +162,24 @@ async def prepare_event_subscriber(
...
 
 
-OutboxCleaner: TypeAlias = Coroutine[Any, Any, None]
+DownloadBucketCleaner: TypeAlias = Coroutine[Any, Any, None]
 
 
 @asynccontextmanager
-async def prepare_outbox_cleaner(
+async def prepare_download_bucket_cleaner(
     *,
```
TheByronHimes: Yielding a coroutine is an unusual approach. Is there a reason this isn't a regular async function that gets awaited from `main.py`?

services/dcs/tests_dcs/test_typical_journey.py
```diff
@@ -17,16 +17,21 @@
...
-from hexkit.providers.s3.testutils import FileObject
+from hexkit.providers.s3.testutils import FileObject, temp_file_object
 from pytest_httpx import HTTPXMock, httpx_mock  # noqa: F401
 
 from dcs.core.models import FileDownloadServed, NonStagedFileRequested
-from tests_dcs.fixtures.joint import CleanupFixture, PopulatedFixture
+from tests_dcs.fixtures.joint import (
+    CleanupFixture,
+    JointFixture,
+    PopulatedFixture,
```
TheByronHimes, line 36: Can delete this comma to get them all on one line

services/dcs/src/dcs/ports/inbound/data_repository.py
```diff
@@ -96,33 +96,32 @@ async def access_drs_object(
...
+        object_storages_config: S3ObjectStoragesConfig,
+        remove_dangling_objects: bool = False,
     ):
-        """Run cleanup task for all outbox buckets configured in the service config."""
-        ...
+        """Run cleanup task for all download buckets configured in the service config."""
 
     @abstractmethod
-    async def cleanup_outbox(self, *, storage_alias: str):
+    async def cleanup_download_bucket(self, *, storage_alias: str):
```
TheByronHimes: This is missing the new `remove_dangling_objects` parameter

services/dcs/tests_dcs/test_typical_journey.py
```diff
@@ -257,3 +262,50 @@ async def test_bucket_cleanup(cleanup_fixture: CleanupFixture, caplog):
     )
 
     assert expected_message in caplog.records[0].message
+
+
+async def test_bucket_cleanup_dangling_object(joint_fixture: JointFixture, caplog):
```
TheByronHimes: We need to check for what happens when there's a non-expired, non-dangling object and `remove_dangling_objects` is True. I suspect it will get deleted when it actually should be left in place.

TheByronHimes (changes requested):
Some things to address first

services/dcs/src/dcs/inject.py
mephenor: There was a reason to do it that way, but let me check if I can't change it to fit our normal approach.

services/dcs/src/dcs/core/data_repository.py
mephenor: Somehow forgot about objects with existing DB entries when moving this here.
I'll set a flag in the except branch that can be used here instead.

mephenor: Addressed in 6755f56afa3ad5712bb15a56b1286f9bcdb2c0ee

services/dcs/tests_dcs/test_typical_journey.py
mephenor: With the changes implemented for the other comment and a more exhaustive test case, this seems to work.
See 6755f56afa3ad5712bb15a56b1286f9bcdb2c0ee for details

services/dcs/src/dcs/inject.py
mephenor: The initial reason no longer exists.
This entrypoint now uses the `data_repository` from `prepare_core` directly.
See 8e1a1710d52f4a140b795f7b4fd3cdf5d58833e8 for details

services/dcs/src/dcs/main.py
```diff
@@ -51,14 +51,17 @@ async def consume_events(run_forever: bool = True):
...
-    """Check if outbox buckets contains files that should be cleaned up and perform clean-up."""
+async def run_download_bucket_cleanup(remove_dangling_objects: bool = False):
+    """Check if download buckets contains files that should be cleaned up and perform clean-up."""
     config = Config()
     configure_logging(config=config)
     configure_opentelemetry(service_name=config.service_name, config=config)
 
-    async with prepare_outbox_cleaner(config=config) as cleanup_outbox:
-        await cleanup_outbox
+    async with prepare_core(config=config) as data_repository:
```
TheByronHimes: It just occurred to me that this will trigger a rebalance on the Kafka broker every time it starts and finishes, because it will start up a Kafka Consumer instance. Can we isolate this logic (maybe in a dedicated class) so it doesn't touch Kafka?

TheByronHimes (changes requested):
Looks better, just one more thing I noticed

services/dcs/src/dcs/main.py
mephenor: Factored out into it's own module in 5c4875e6bfa77ba2d19eb62964becb2a21a316a0.

TheByronHimes (approved):

PR #159, opened 2026-03-09
UCS: Sarcastic Fringehead (GSI-2079)

TheByronHimes:
Bumps UCS version to `11.0.0`

### Changes:
**Replaced FileUploadReport**: The outdated concept of a `FileUploadReport` has been replaced with the new `InterrogationSuccess`/`InterrogationFailure` concepts. The same goes for the previously implemented code which handled FileUploadReports - we now have two event consumption points. 

**Local schema definitions**: Like other repo changes, I've locally defined many of the airborne schemas temporarily until things have gelled. I don't want to burn through several dead major-version `ghga-event-schemas` releases just because dev/testing revealed something we didn't catch beforehand. So far this has proved to be a good move.

**FileUploadBox** versioning: Like the `ResearchDataUploadBox` in the UOS, the FUB now has a `version` attribute (`int`). The way it works in the UCS is straightforward: anything that changes the FUB **_directly_** causes the version to increment. Requests to modify the box require a version number to make sure that the request is not operating on outdated information. This is also how it works in the UOS.

**`S3Client`**: The main `UploadController` class was bloated from all the S3 logic, so I extracted that into a new outbound adapter called the `S3Client`. This is consistent with the DHFS, which also had to perform a lot of S3-related error handling and fallback operations. 

**`FileUpload` Attributes**: The `FileUpload` schema now has more attributes than in the initial UCS implementation. For example, `size` was split out into `decrypted_size` and `encrypted_size`, and we now track `object_id` too. We also require some of this information up front instead of upon upload completion.

---

**Outstanding**: The UCS still needs some final work to guard against runaway uploads (intentional or otherwise).


services/ucs/src/ucs/main.py
```diff
@@ -66,14 +68,11 @@ async def consume_events(run_forever: bool = True):
         await event_subscriber.run(forever=run_forever)
 
 
-# TODO: In the actual UCS ticket, move index creation to the .get_dao() call
 async def initialize() -> None:
     """Operations to be run in an init container before service startup."""
     config = Config()
-
     configure_logging(config=config)
     configure_opentelemetry(service_name=config.service_name, config=config)
```
mephenor: If there's nothing in here, only the log message, then this shouldn't be run.

services/ucs/src/ucs/adapters/inbound/fastapi_/http_exceptions.py
```diff
@@ -52,10 +52,33 @@ def __init__(self, *, box_id: UUID4, status_code: int = 404):
...
-    exception_id = "lockedBox"
+    exception_id = "boxStateError"
+
+    class DataModel(BaseModel):
+        """Model for exception data"""
+
+        box_id: UUID4
+        box_state: str
+
+    def __init__(self, *, box_id: UUID4, box_state: str, status_code: int = 409):
```
mephenor: `box_state` should be `models.UploadBoxState` instead

services/ucs/src/ucs/core/controller.py
```diff
@@ -54,78 +54,115 @@ def __init__(
...
+
+            extra = {  # only for logging
+                "box_id": box.id,
+                "file_alias": alias,
+                "old_file_upload_id": existing.id,
+                "old_state": existing.state,
+                "new_file_upload_id": file_upload.id,
+            }
+
+            # Examine the existing FileUpload - if it's failed cancelled, we can replace
```
mephenor:
```suggestion
            # Examine the existing FileUpload - if it's failed or cancelled, we can replace
```

services/ucs/src/ucs/core/controller.py
```diff
@@ -54,78 +54,115 @@ def __init__(
...
+                    "Cleaned out S3UploadDetails for old, %s FileUpload %s because it"
+                    + " is being superseded by FileUpload %s.",
+                    existing.state,
+                    existing.id,
+                    file_upload.id,
+                    extra=extra,
+                )
+            await self._file_upload_dao.delete(existing.id)
+            await self._file_upload_dao.insert(file_upload)
+            return file_upload
```
mephenor: That's a long except clause with other different potential error sources involved.

This should at least be factored out for readability.

services/ucs/src/ucs/core/controller.py
```diff
@@ -136,8 +173,8 @@ async def _get_unlocked_box(self, *, box_id: UUID4) -> FileUploadBox:
             raise error from err
 
         # Verify that the box is not locked
```
mephenor:
```suggestion
        # Verify that the box is not locked or archived
```

services/ucs/tests_ucs/integration/test_controller.py
```diff
@@ -221,7 +233,11 @@ async def test_s3_upload_completed_but_db_not_updated(joint_fixture: JointFixtur
     async with set_correlation_id(uuid4()):
         box_id = await controller.create_file_upload_box(storage_alias="test")
         file_id = await controller.initiate_file_upload(
-            box_id=box_id, alias="test-file", size=1024
+            box_id=box_id,
+            alias="test-file",
+            decrypted_size=1024,
+            encrypted_size=1024,
```
mephenor: Business logic doesn't care `encrypted_size` being slightly larger in reality?

services/ucs/src/ucs/core/controller.py
```diff
@@ -219,109 +216,101 @@ async def _remove_incomplete_file_upload(
...
-            )
-            log.critical(str(error), exc_info=True, extra=extra)
+            #  then there cannot be an upload for said file_id. Since each FileUpload
+            #  gets a freshly generated object_id, a collision here would be extremely
+            #  unlikely. The most likely cause is a crash between creating the S3 upload
+            #  and inserting the S3UploadDetails. We can't assign S3 upload IDs, so if
+            #  that data isn't saved to the DB, it is only preserved in the logs. There
+            #  is no straightforward way to get the upload ID programmatically, so we
+            #  can't auto-abort it, either. In this case a developer will have to
+            #  manually intervene to cancel the upload. We will delete the FileUpload,
```
mephenor: Not sure that's something a dev should do if we are talking about ingress buckets, that's more on the ops side of things either way and can probably be solved by some auto cleanup/bucket policy.

services/ucs/src/ucs/core/controller.py
```diff
@@ -451,116 +456,91 @@ async def complete_file_upload(
...
-            await self._update_box_stats(box=box)
             await self._compare_checksums(
-                object_storage=object_storage,
-                bucket_id=bucket_id,
-                file_id=file_id,
+                s3_upload_details=s3_upload_details,
+                file_upload=file_upload,
                 expected_checksum=encrypted_checksum,
             )
+            await self._update_box_stats(box_id=box_id, version=box_version)
```
mephenor: Need some local context here on how this state would be reached. Can this happen without this method being called a second time?

Naively these two calls should have already been executed by that time in the code below. 
Is `self._update_box_stats` idempotent in that context?

services/ucs/src/ucs/core/controller.py
```diff
@@ -451,116 +456,91 @@ async def complete_file_upload(
...
+        file_upload.inbox_upload_completed = True
+        file_upload.state_updated = now_utc_ms_prec()
         s3_upload_details.completed = now_utc_ms_prec()
         await self._file_upload_dao.update(file_upload)
         await self._s3_upload_details_dao.update(s3_upload_details)
 
         # Update the FileUploadBox with new size and file count
-        await self._update_box_stats(box=box)
+        await self._update_box_stats(box_id=box_id, version=box_version)
         log.debug("DB data updated for upload completion of file %s", file_id)
```
mephenor: Can be log.info instead, if there's no other log message at point in the calling code.

services/ucs/src/ucs/core/controller.py
```diff
@@ -451,116 +456,91 @@ async def complete_file_upload(
...
         box = await self._get_unlocked_box(box_id=box_id)
+        box_version = box.version
 
         # Retrieve the FileUpload data
         try:
             file_upload = await self._file_upload_dao.get_by_id(file_id)
         except ResourceNotFoundError:
             log.info("File %s not found - presumed already deleted.", file_id)
+            # Call _update_box_stats() to be consistent with complete_file_upload()
+            await self._update_box_stats(box_id=box_id, version=box_version)
```
mephenor: Similar question to the above method:
Wouldn't this have already been taken care of by the code below or is there another way to reach this state?

services/ucs/src/ucs/core/controller.py
```diff
@@ -653,12 +674,94 @@ async def unlock_file_upload_box(self, *, box_id: UUID4) -> None:
...
+                or not file.encrypted_parts_sha256
+                or len(file.encrypted_parts_md5) != len(file.encrypted_parts_sha256)
+            ):
+                raise self.FileArchivalError(
+                    f"File part checksums appear corrupted for file {file.id}."
+                )
+            elif file.failure_reason:
+                raise self.FileArchivalError(
+                    f"The 'failure_reason' for file {file.id} is unexpectedly filled out."
+                )
```
mephenor, line 773: This will raise on the first such file encountered.
What are the chances there is more than one?

services/ucs/src/ucs/core/controller.py
```diff
@@ -653,12 +674,94 @@ async def unlock_file_upload_box(self, *, box_id: UUID4) -> None:
...
+            file.state = "awaiting_archival"
+            file.state_updated = now_utc_ms_prec()
+            await self._file_upload_dao.update(file)
```
mephenor, line 776: We don't have functionality in the hexkit DAO to do batch inserts/updates, correct?

services/ucs/src/ucs/core/controller.py
```diff
@@ -678,25 +781,82 @@ async def get_box_file_info(self, *, box_id: UUID4) -> list[FileUpload]:
         file_uploads = [
             x
             async for x in self._file_upload_dao.find_all(
-                mapping={"box_id": box_id, "completed": True}
+                mapping={"box_id": box_id, "inbox_upload_completed": True}
             )
         ]
         file_uploads.sort(key=lambda x: x.alias)
```
mephenor, line 805: Another point of thinking about DB primitives: 
We can't sort in the find_all call?

services/ucs/src/ucs/core/controller.py
```diff
@@ -678,25 +781,82 @@ async def get_box_file_info(self, *, box_id: UUID4) -> list[FileUpload]:
...
+
+        # Attempt to delete S3 file even if this event has been processed before
+        try:
+            s3_upload_details = await self._s3_upload_details_dao.get_by_id(file_id)
+        except ResourceNotFoundError as err:
+            error = self.S3UploadDetailsNotFoundError(file_id=file_id)
+            log.error(error, extra={"file_id": file_id})
+            raise error from err
+
+        await self._remove_completed_file_upload(s3_upload_details=s3_upload_details)
```
mephenor, line 864: When are the s3_upload_details removed, i.e. won't this potentially raise on reprocessing instead of just running through?

services/ucs/src/ucs/core/controller.py
mephenor, line 915: Previous question also applies to this code.

services/ucs/tests_ucs/unit/test_s3client.py
```diff
@@ -0,0 +1,379 @@
...
+        initiated=now_utc_ms_prec(),
+    )
+
+
+@pytest.fixture()
+def patch_s3_calls(monkeypatch):
+    """Mocks the object storage provider with an InMemObjectStorage object"""
+    pass
+    monkeypatch.setattr(
+        f"{InMemS3ObjectStorages.__module__}.S3ObjectStorage", InMemObjectStorage
```
mephenor, line 86: Is this pointing to the right module?

services/ucs/tests_ucs/unit/test_s3client.py
```diff
@@ -0,0 +1,379 @@
...
+        object_id=object_id,
+        s3_upload_id=s3_upload_id,
+        initiated=now_utc_ms_prec(),
+    )
+
+
+@pytest.fixture()
+def patch_s3_calls(monkeypatch):
+    """Mocks the object storage provider with an InMemObjectStorage object"""
+    pass
```
mephenor: ?

services/ucs/tests_ucs/unit/test_s3client.py
```diff
@@ -0,0 +1,379 @@
...
+        decrypted_size=1000,
+        encrypted_size=1100,
+        part_size=100,
+    )
+
+
+def make_s3_upload_details(
+    *,
+    file_id: UUID4 | None = None,
+    s3_upload_id: str = "",
```
mephenor: Might be more meaningful to provide a non-empty default string to spot the `uninitialized` state easier.

services/ucs/tests_ucs/unit/test_s3client.py
mephenor, line 1: Some of the tests follow similar patterns and could potentially condensed by converting them to parametrized test cases with a bit of extra logic.

services/ucs/tests_ucs/unit/test_controller.py
```diff
@@ -115,14 +55,18 @@ async def test_create_new_file_upload(rig: JointRig):
 
     # Then create a FileUpload within the box
     file_id = await controller.initiate_file_upload(
-        box_id=box_id, alias="test_file", size=1024
+        box_id=box_id,
+        alias="test_file",
+        decrypted_size=1024,
+        encrypted_size=1024,
```
mephenor: As mentioned in the other comment, encrypted_size should probably be larger than the decrypted size.

Could factor out those common values in the test files, so they are only defined in one place.

mephenor (changes requested):

services/ucs/src/ucs/core/controller.py
TheByronHimes: No. Hexkit obscures that because it performs the iteration itself instead of passing us the cursor.

services/ucs/src/ucs/core/controller.py
TheByronHimes: Correct.

services/ucs/src/ucs/main.py
TheByronHimes: Deleted, though I would have rather left it in as a ready receptacle.

services/ucs/src/ucs/core/controller.py
TheByronHimes: `self._update_box_stats()` is by design not idempotent because of the version increment. This branch is the replay branch. If two calls come in to complete the same file upload (maybe a network blip, maybe a connector bug, maybe a flawed modification for the connector), the second call would hit here. But as I'm thinking about it, this would raise an error if the second call carried the same box version as the first call, which would probably result in a file deletion. So I'll actually take the comparison out of this branch and merely log/return.

services/ucs/src/ucs/core/controller.py
TheByronHimes: Yeah, and the original thinking was that if something prevented the stats update the first time around, this would get it caught up. But maybe that's not the way to do it.

services/ucs/tests_ucs/unit/test_s3client.py
TheByronHimes: Yeah, it's supposed to patch the locally defined multinode storage mock so that when it uses `S3ObjectStorage` it actually uses `InMemObjectStorage`.

services/ucs/tests_ucs/unit/test_s3client.py
TheByronHimes: I've done a little bit of that in [ea7df40](https://github.com/ghga-de/file-services-backend/pull/159/commits/ea7df409074b77df8aa64cfbad1f00d4cb2037ff) and [d87fa77](https://github.com/ghga-de/file-services-backend/pull/159/commits/d87fa77e58a6123e394a0309f810af0c03fc02ff) but like having them split out. When a test fails it's easier to diagnose when the test is more narrowly targeted.

services/ucs/src/ucs/core/controller.py
TheByronHimes: Please take a look at [dcd94e3](https://github.com/ghga-de/file-services-backend/pull/159/commits/dcd94e3a8eeb40e0efb8370a2aa1516ec93344ae) and see if that is an improvement

services/ucs/tests_ucs/integration/test_controller.py
TheByronHimes: No. I can make that a check though.

TheByronHimes: In [1d5b8a8](https://github.com/ghga-de/file-services-backend/pull/159/commits/1d5b8a822e55ba80d449f1ed0894235e3cbf3f55) I added a validator to do that as part of HTTP request payload validation

services/ucs/tests_ucs/unit/test_controller.py
TheByronHimes: I've made the sizes more realistic and added constants in the fixtures/utils module.

services/ucs/src/ucs/core/controller.py
TheByronHimes: This now logs and returns without calling `self._update_box_stats()` again

services/ucs/src/ucs/core/controller.py
TheByronHimes: Hard to say, because it's not supposed to happen in the first place.

services/ucs/src/ucs/core/controller.py
TheByronHimes: I don't know why that comment was there. They shouldn't be deleted.

services/ucs/src/ucs/core/controller.py
mephenor: Ok, then let's keep it as is.
Wasn't sure if it would make more sense to collect all information first and then raise

services/ucs/src/ucs/adapters/inbound/fastapi_/rest_models.py
```diff
@@ -60,9 +60,35 @@ class FileUploadCreationRequest(BaseModel):
...
+class FileUploadCreationResponse(BaseModel):
+    """Response body for newly created FileUploads"""
+
+    file_id: UUID4 = Field(
+        ..., description="The UUID4 identifier assigned to the FileUpload"
+    )
+    alias: str = Field(
+        ...,
+        description="The alias for the file within the box (must be unique within the box)",
+    )
```
mephenor, line 86: I thought the alias was user provided and exchanged for the ID?
Is this just returned as is for client side validation?

services/ucs/tests_ucs/unit/test_api.py
```diff
@@ -204,6 +204,11 @@ async def test_create_file_upload_endpoint_auth(
     )
     response = await rest_client.post(url, json=body, headers=good_token_header)
     assert response.status_code == 201
+    # assert response.json() == {
+    #     "file_id": TEST_FILE_ID,
+    #     "alias": body["alias"],
+    #     "storage_alias": "HD01",
+    # }
```
mephenor: Is this relevant?

services/ucs/src/ucs/adapters/inbound/fastapi_/rest_models.py
```diff
@@ -60,9 +60,35 @@ class FileUploadCreationRequest(BaseModel):
...
+
+    @model_validator(mode="after")
+    def encrypted_size_exceeds_decrypted_size(self) -> "FileUploadCreationRequest":
+        """Ensure encrypted_size is larger than decrypted_size."""
+        if self.encrypted_size <= self.decrypted_size:
+            raise ValueError(
+                f"encrypted_size ({self.encrypted_size}) must be larger than"
+                f" decrypted_size ({self.decrypted_size})"
+            )
+        return self
```
mephenor, line 72: Is it transparent in the response, why this fails?

services/ucs/tests_ucs/unit/test_api.py
TheByronHimes: It sure is, looks like I forgot to uncomment it.

services/ucs/src/ucs/adapters/inbound/fastapi_/rest_models.py
TheByronHimes: Exactly. This enables the client to verify that the response corresponds to the expected request. Just cheap validation

services/ucs/src/ucs/adapters/inbound/fastapi_/rest_models.py
TheByronHimes: As clear as any pydantic-sourced validation error. The error details are communicated in the response's content and text properties, as well as the body. The likelihood of this specific error being triggered is the same as any other model violation, imo.

mephenor (approved):

PR #160, opened 2026-03-16
FIS: Log user agent headers for received requests (GSI-2226)

TheByronHimes:
This logs the received user-agent headers along with the data hub alias (storage_alias) for FIS's HTTP API.

### DIFF
--- a/services/fis/src/fis/adapters/inbound/fastapi_/routes.py
+++ b/services/fis/src/fis/adapters/inbound/fastapi_/routes.py
@@ -17,7 +17,7 @@
 import logging
 from typing import Annotated
 
-from fastapi import APIRouter, Body, HTTPException, status
+from fastapi import APIRouter, Body, HTTPException, Request, status
 from pydantic import UUID4
 
 from fis.adapters.inbound.fastapi_ import dummies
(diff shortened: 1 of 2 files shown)


services/fis/src/fis/adapters/inbound/fastapi_/routes.py
```diff
@@ -30,15 +30,45 @@
...
+    request: Request, endpoint_description: str, storage_alias: str = ""
+) -> None:
+    """Log a structured message from the User-Agent header.
+
+    Expects the format 'ClientName/1.0.0'. Adapts the message for a missing or
+    malformed header, or a missing storage alias.
+    """
+    raw_user_agent_header = request.headers.get("user-agent")
+    extra = {"user_agent": raw_user_agent_header}
+    client_name, _, client_version = (raw_user_agent_header or "").partition("/")
```
dontseyit: I don't think this is real concern at the moment but some libraries build user-agents from dependencies as well, e.g. Boto3 has a collection of clients. It ends up something like "Boto3/1.2.3 Python/3.12 Botocore/1.2.3 myApp/1.2.3". We probably don't expect such a request arriving to FIS, but `partition("/")` gets only 'Boto3' as client_name, and the version will be the rest. I guess at worst we get a weird-looking log.

TheByronHimes: I've adjusted it so that it will only try to interpret name and version if there's only one occurrence of "/". Otherwise it will just use whatever string was in the header. I think regex or verifying the version with the semver library is overkill for this.

dontseyit (approved):

PR #161, opened 2026-03-17
All Services: Use latest schemas lib (GSI-2216)

TheByronHimes:
The PCS hadn't been updated yet, so I bumped it to version `6.0.0`


services/fis/openapi.yaml
```diff
@@ -125,8 +125,7 @@ components:
           title: Reason
         secret:
           anyOf:
-          - contentMediaType: application/octet-stream
-            format: password
```
mephenor, line 129: What changed at the boundary here?

mephenor (approved):

services/fis/openapi.yaml
TheByronHimes: I don't know, it seems to be reverted after re-running the update script

PR #162, opened 2026-03-18
UCS: Handle FileInternallyRegistered events (GSI-2229)

TheByronHimes:
Adds support for FileInternallyRegistered and FileDeletionRequested events

### DIFF
--- a/services/ucs/src/ucs/core/controller.py
+++ b/services/ucs/src/ucs/core/controller.py
@@ -20,7 +20,11 @@ from contextlib import suppress
 from typing import Any
 from uuid import uuid4
 
-from ghga_event_schemas.pydantic_ import InterrogationFailure, InterrogationSuccess
+from ghga_event_schemas.pydantic_ import (
+    FileInternallyRegistered,
+    InterrogationFailure,
+    InterrogationSuccess,
+)
 from hexkit.protocols.dao import NoHitsFoundError, UniqueConstraintViolationError
 from hexkit.utils import now_utc_ms_prec
 from pydantic import UUID4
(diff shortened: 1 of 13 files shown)


services/ucs/src/ucs/core/controller.py
```diff
@@ -908,3 +912,82 @@ async def process_interrogation_failure(
...
+            log.info(
+                "FileUpload %s is already marked '%s', further action presumed unnecessary.",
+                file_id,
+                file_upload.state,
+            )
+            return
+
+        # This will result in a second FileUpload fetch, but alternative is to
+        #  replicate removal logic here
+        await self.remove_file_upload(box_id=file_upload.box_id, file_id=file_id)
```
mephenor, line 993: Could also change the signature to take a boolean and perform the lookup conditionally, though that muddies the logic a bit.

services/ucs/src/ucs/core/controller.py
```diff
@@ -908,3 +912,82 @@ async def process_interrogation_failure(
...
         await self._remove_completed_file_upload(s3_upload_details=s3_upload_details)
+
+    async def process_internal_file_registration(
+        self, *, registration_metadata: FileInternallyRegistered
+    ) -> None:
+        """Update a FileUpload state to 'archived' and verify other data.
+
+        Raises:
+        - `FileUploadNotFound` if the FileUpload isn't found.
+        - `FileUploadStateError` if the FileUpload's details aren't what's expected.
```
mephenor, line 923: Those events will land in the DLQ, correct?

services/ucs/tests_ucs/fixtures/utils.py
```diff
@@ -137,3 +141,36 @@ def delete_file_token_header(
...
+        part_size=PART_SIZE,
+    )
+
+    if state != "init":
+        file_upload.decrypted_sha256 = "my-decrypted-sha"
+        file_upload.encrypted_parts_md5 = ["a1", "b2", "c3"]
+        file_upload.encrypted_parts_sha256 = ["a1", "b2", "c3"]
+
+    if state not in ["init", "inbox"]:
+        file_upload.secret_id = "the-secret-is"
```
mephenor, line 174: We might never know ;)

services/ucs/tests_ucs/integration/test_controller.py
```diff
@@ -631,3 +631,62 @@ async def test_file_upload_report_happy(joint_fixture: JointFixture):
...
+    await joint_fixture.event_subscriber.run(forever=False)
+
+    # Publish/Consume a deletion request for a file that doesn't exist (no error)
+    unknown_event = FileDeletionRequested(file_id=uuid4())
+    await joint_fixture.kafka.publish_event(
+        payload=unknown_event.model_dump(mode="json"),
+        type_=config.file_deletion_request_type,
+        topic=config.file_deletion_request_topic,
+    )
+    await joint_fixture.event_subscriber.run(forever=False)
```
mephenor: Both of these only log?
Could add caplog and check message content to verify the correct path was taken

services/ucs/tests_ucs/unit/test_controller.py
```diff
@@ -1171,3 +1181,94 @@ async def patched_insert(dto):
...
+        )
+
+    # Now insert the FileUpload and run the event handling function
+    await rig.file_upload_dao.insert(file_upload)
+
+    await sleep(0.1)  # sleep so that timestamp comparisons are valid
+    await rig.controller.process_internal_file_registration(registration_metadata=event)
+    updated_file_upload = await rig.file_upload_dao.get_by_id(file_upload.id)
+    assert updated_file_upload.state == "archived"
+    assert updated_file_upload.state_updated >= file_upload.state_updated
```
mephenor: \>= seems like a typo

mephenor (changes requested):

services/ucs/src/ucs/core/controller.py
TheByronHimes: That's right

services/ucs/tests_ucs/unit/test_controller.py
TheByronHimes: Yeah, should be `>`

services/ucs/src/ucs/core/controller.py
TheByronHimes: Yeah, and it needs to get the FileUpload too in order to reference certain fields. This method is called from the HTTP routes as well. _At first glance_ there doesn't seem to be a clean way to segment the logic without some rippling refactoring.

services/ucs/tests_ucs/integration/test_controller.py
TheByronHimes: Can do and done

mephenor (approved):

PR #163, opened 2026-03-27
[minor ver bump] UCS: Fold S3UploadDetails into FileUpload for simpler code (GSI-2253)

TheByronHimes:
Removes the `S3UploadDetails` model completely from UCS, folding the important fields instead into the `FileUpload` model (mainly upload_id).
This simplifies some of the code (one less DAO), reduces the number of different failure points, consolidates some error handling, and is more in line with best practices for NoSQL data modeling.
Bumps the minor version so UCS is now at `11.1.0` and the monorepo itself goes to `10.1.0`

### DIFF
--- a/services/ucs/src/ucs/core/controller.py
+++ b/services/ucs/src/ucs/core/controller.py
@@ -16,7 +16,6 @@
 """Implements the UploadController class to manage file uploads"""
 
 import logging
-from contextlib import suppress
 from typing import Any
 from uuid import uuid4
 
@@ -30,13 +29,12 @@ from hexkit.utils import now_utc_ms_prec
 from pydantic import UUID4
 
 from ucs.config import Config
-from ucs.core.models import FileUpload, FileUploadBox, S3UploadDetails
+from ucs.core.models import FileUpload, FileUploadBasics, FileUploadBox
 from ucs.ports.inbound.controller import UploadControllerPort
 from ucs.ports.outbound.dao import (
     FileUploadBoxDao,
     FileUploadDao,
     ResourceNotFoundError,
-    S3UploadDetailsDao,
 )
 from ucs.ports.outbound.storage import S3ClientPort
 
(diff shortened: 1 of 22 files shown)


services/ucs/src/ucs/core/models.py
```diff
@@ -20,16 +20,18 @@
...
+    alias: str
     storage_alias: str
-    s3_upload_id: str
-    initiated: UTCDatetime
-    completed: UTCDatetime | None = None
+    bucket_id: str
+    object_id: UUID4
+    decrypted_size: int
+    encrypted_size: int
+    part_size: int
```
mephenor: What about some descriptions + will this be factored out into a baseclass for `FileUpload` in ghga-event-schemas?

services/ucs/tests_ucs/unit/test_s3client.py
```diff
@@ -15,42 +15,26 @@
...
-        file_id=file_id,
-        storage_alias=TEST_STORAGE_ALIAS,
-        bucket_id=TEST_BUCKET,
-        object_id=object_id,
-        s3_upload_id=s3_upload_id,
-        initiated=now_utc_ms_prec(),
+def _to_basics(file_upload: FileUpload) -> FileUploadBasics:
+    """Extract a FileUploadBasics from a FileUpload."""
+    return FileUploadBasics(
+        **{k: getattr(file_upload, k) for k in FileUploadBasics.model_fields}
```
mephenor: Could use a more descriptive variable name here

services/ucs/src/ucs/core/controller.py
```diff
@@ -292,60 +242,59 @@ async def initiate_file_upload(
         extra["storage_alias"] = storage_alias
         extra["bucket_id"] = bucket_id
 
-        initiated = now_utc_ms_prec()  # Generate timestamp early to minimize error risk
-        file_upload = await self._insert_file_upload_if_new(
-            box=box,
+        file_id = uuid4()
+        object_id = uuid4()
```
mephenor, line 246: Where were those populated previously?

TheByronHimes: Inside of `_insert_file_upload()`, which used to be called `_insert_file_upload_if_new()`

mephenor (approved):

PR #164, opened 2026-04-21
FIS: Don't require decrypted_sha256 during init state

TheByronHimes:
Set `BaseFileInformation.decrypted_sha256` to optional when state is either `"init"`, `"failed"`, or `"cancelled"`.

Bumps version to `10.0.1`


dontseyit (approved):
Looks good to me

PR #165, opened 2026-04-22
EKSS: Expect secret only (GSI-2322)

TheByronHimes:
Bumps EKSS to version `5.0.0`.  
Bumps the monorepo to version `11.0.0`

This is the Sarcastic Fringehead update for EKSS. EKSS has 3 endpoints, and only one of them (`POST /secrets`) has changed

Before: 
- Endpoint requires submitter pub key and the first part of the file containing the envelope
  - While the submitter pub key was supplied to envelope extraction fn, it doesn't seem that it actually had any effect.
- Extracts original file secret and content offset from envelope
- Generates new random file secret --> stores it in vault
- Returns original file secret, new file secret, vault ID of new secret, and file content offset

Now:
- Endpoint requires body to contain base64-encoded encrypted file secret created by DHFS (and sent via FIS).
  - To be clear, this is the secret used to re-encrypt the interrogated file content
- Decodes and then decrypts the encrypted file secret to obtain the base64-encoded secret
- Decodes the decrypted secret to get the raw file secret --> stores it in vault
- Returns vault ID of the stored secret.

The endpoints for secret deletion and envelope retrieval are basically unchanged, because the way secrets are handled downstream of the vault insertion is identical.

*Structure*  
The business logic was extracted from the HTTP router module and placed in a new `SecretsHandler` core class. The `VaultAdapter` was renamed to `VaultClient` and given a port definition. The actual logic for everything is more or less the same as it was before, but updated for triple hex architecture. Thus, the VaultClient logic is unchanged. The logic now contained in the SecretsHandler came from the router module and some standalone logic that used to be in the core.

*Testing*  
Testing reflects the same structural adaptations. There are unit tests for the HTTP API, SecretsHandler, and VaultClient. There is an integration test module that checks the happy path of all three endpoints with a vault testcontainer, and this might have some overlap with the unit tests. There were some fixtures that are no longer necessary & have been removed, and the crytp4gh keypair generation has been improved somewhat. 

*Other*  
FIS needs to be updated so that it expects the response payload to be {"secret_id": ...} instead of a raw string.  There is a [PR for that](https://github.com/ghga-de/file-services-backend/pull/166).
DHFS uses `base64.urlsafe_b64encode` to encode the raw file secret bytes as a mistake on my part, which is why EKSS uses `base64.urlsafe_b64decode` too. The `urlsafe_` isn't necessary since the encoded string is anyway wrapped up in a layer of encryption and encoding, but the two need to be aligned. I'm inclined to leave it as-is, but we could alternatively change it to the non-`urlsafe_` version and release a new major version of DHFS.


mephenor:
We can keep the `urlsafe_` version, as long as the corresponding sides align

services/ekss/src/ekss/adapters/inbound/fastapi_/routes.py
```diff
@@ -65,6 +43,22 @@
         ),
         "model": exceptions.HttpDecodingError.get_body_model(),
     },
+    "decryptionError": {
+        "description": "Could not decrypt the submitted file secret",
+        "model": exceptions.HttpDecryptionError.get_body_model(),
+    },
+    "envelopeCreationError": {
+        "description": "Could not create envelope for the requested secret.",
```
mephenor:
```suggestion
        "description": "Could not create an envelope for the requested secret.",
```

services/ekss/src/ekss/adapters/inbound/fastapi_/routes.py
```diff
@@ -81,67 +75,48 @@ async def health():
...
     },
 )
 @TRACER.start_as_current_span("routes.post_encryption_secret")
 async def post_encryption_secret(
     *,
-    envelope_query: models.InboundEnvelopeQuery,
-    config: Config = Depends(config_injector),
+    body: Annotated[
+        bytes,
+        Body(
```
mephenor, line 95: How is the body payload modeled here, just a string?
Also, we have fixed size secrets, so the encoding should theoretically also have a fixed size and we could early reject invalid secrets that way.

services/ekss/src/ekss/adapters/inbound/fastapi_/routes.py
mephenor, line 1: Can the secret be leaked via stacktrace here (for any/all of the except branches)?

services/ekss/src/ekss/ports/inbound/secrets.py
mephenor, line 1: The `...` can be removed from every abstractmethod here.

services/ekss/tests_ekss/fixtures/__init__.py
mephenor, line 1: Where's the docstring, Lebowski?

services/ekss/src/ekss/core/secrets.py
```diff
@@ -0,0 +1,119 @@
...
+        try:
+            file_secret = self._vault_client.get_secret(key=secret_id)
+        except VaultClientPort.SecretRetrievalError as err:
+            error = self.SecretRetrievalError(secret_id=secret_id)
+            log.error(error, extra={"secret_id": secret_id})
+            raise error from err
+
+        try:
+            keys = [(0, self._private_key, client_pubkey)]
+            header_content = crypt4gh.header.make_packet_data_enc(0, file_secret)
```
mephenor, line 59: Can secret bytes be leaked in the reraised error stacktrace here?

services/ekss/src/ekss/core/secrets.py
```diff
@@ -0,0 +1,119 @@
...
+        """
+        try:
+            base64_file_secret = decrypt(encrypted_secret, self._private_key)
+        except Exception as err:
+            error = self.SecretDecryptionError()
+            log.error(error)
+            raise error from err
+
+        try:
+            file_secret = base64.urlsafe_b64decode(base64_file_secret)
```
mephenor, line 87: Same concern about the secret bytes here

services/ekss/src/ekss/core/secrets.py
```diff
@@ -0,0 +1,119 @@
...
+        except Exception as err:
+            error = self.SecretDecodeError()
+            log.error(error)
+            raise error from err
+
+        try:
+            secret_id = self._vault_client.store_secret(secret=file_secret)
```
mephenor, line 94: And here

services/ekss/src/ekss/ports/outbound/vault.py
```diff
@@ -0,0 +1,59 @@
...
+        """Raised when a secret cannot be deleted from the vault"""
+
+    @abstractmethod
+    def get_secret(self, *, key: str) -> bytes:
+        """Retrieve a secret at the subpath of the given prefix denoted by key.
+
+        Key should be a UUID4 returned by store_secret on insertion.
+
+        Raises: SecretRetrievalError: if the secret cannot be retrieved.
+        """
```
mephenor, line 44: Do we want to enforce the key format here and reject early?

services/ekss/tests_ekss/integration/test_api_integrated.py
```diff
@@ -0,0 +1,96 @@
...
+    ):
+        response = await client.get(url=f"/secrets/{secret_id}/envelopes/{client_pk}")
+        assert response.status_code == 200
+        body = response.json()
+
+    # Inspect the response body
+    content = base64.b64decode(body["content"])
+    assert content
+
+    # Extract the secret from the envelope using the user's pub key and verify it matches
```
mephenor: This uses the private key, i.e. the code directly below the comment

services/ekss/tests_ekss/unit/test_api.py
```diff
@@ -15,20 +15,148 @@
...
+    assert response.json() == {
+        "content": base64.b64encode(fake_envelope).decode("utf-8")
+    }
+
+
+async def test_post_secret(client_fixture: ClientFixture, keypair: KeypairFixture):
+    """Test the POST /secrets endpoint (no core errors)"""
+    # Should get an error when body is empty
+    response = await client_fixture.client.post("/secrets")
+    assert response.status_code == 422
```
mephenor, line 58: What about an existing, but empty body to check the min-length requirement, so content=""?

mephenor (changes requested):

services/ekss/src/ekss/adapters/inbound/fastapi_/routes.py
TheByronHimes: Yes, just a string. It now has the right type-hint and a fixed length requirement of 124 bytes.

services/ekss/src/ekss/adapters/inbound/fastapi_/routes.py
TheByronHimes: This isn't public facing and logs are also treated as sensitive. We can switch everything to `Secret-` in a follow-up PR maybe?

services/ekss/tests_ekss/unit/test_api.py
TheByronHimes: Added

services/ekss/src/ekss/ports/outbound/vault.py
TheByronHimes: We don't have control over what Vault generates so let's keep loose. I removed the UUID4 references.

services/ekss/src/ekss/constants.py
```diff
@@ -18,3 +18,4 @@
 
 SERVICE_NAME = "ekss"
 TRACER = trace.get_tracer_provider().get_tracer(SERVICE_NAME)
+ENCODED_ENCRYPTED_KEY_SIZE = 124  # length of base64-encoded crypt4gh-encrypted secret
```
mephenor, line 21: How is that calculated, 44 from encoding + 80 from encryption?

TheByronHimes: Appears so. I had derived it by generating the encrypted payload twenty thousand times to verify that it was always the same. Looking at the steps, it's 44 after encoding, 124 after calling `encrypt()`

mephenor (approved):

PR #166, opened 2026-04-22
FIS: Update to expect correct EKSS response format

TheByronHimes:
Goes along with the [EKSS PR](https://github.com/ghga-de/file-services-backend/pull/165).

Bumps FIS version to `11.0.0`

Monorepo version was bumped in the EKSS PR.

### DIFF
--- a/services/fis/src/fis/adapters/outbound/secrets.py
+++ b/services/fis/src/fis/adapters/outbound/secrets.py
@@ -79,7 +79,12 @@ class SecretsClient(SecretsClientPort):
             )
             raise self.SecretsApiError()
 
-        return response.json()
+        try:
+            secret_id = response.json().get("secret_id")
+        except Exception as err:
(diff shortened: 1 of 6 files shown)


services/fis/src/fis/adapters/outbound/secrets.py
```diff
@@ -79,7 +79,12 @@ async def deposit_secret(self, *, secret: SecretBytes) -> str:
             )
             raise self.SecretsApiError()
 
-        return response.json()
+        try:
+            secret_id = response.json().get("secret_id")
+        except Exception as err:
+            raise self.SecretsApiError() from err
```
mephenor, line 86: This should also provide a corresponding log message, e.g. `failed retrieving secret_id from response`

mephenor (changes requested):

mephenor (approved):

PR #170, opened 2026-05-04
UCS: Box size limits (GSI-2336)

TheByronHimes:
This adds support for specifying a `max_size` (a requirement actually) for file upload boxes.

The value can be adjusted through the update endpoint -- RS will sign a WOT with work type `"resize"`. But if you do this and set the value to something lower than the current `box.size`, it will trigger a `409` response. Internally this is a new error on the UploadControllerPort.

If a user tries to upload a new file that would exceed the box size limit, the response will be a 507 Insufficient Storage error.

The size


services/ucs/src/ucs/adapters/inbound/fastapi_/http_exceptions.py
```diff
@@ -234,6 +234,79 @@ def __init__(self, *, file_id: UUID4, status_code: int = 400):
         )
 
 
+class HttpBoxMaxSizeBelowCurrentSizeError(HttpCustomExceptionBase):
```
Cito: name is a bit long - how about naming it HttpBoxMaxSizeCannotShrinkError or HttpBoxShrinkNotAllowedError?

services/ucs/src/ucs/adapters/inbound/fastapi_/http_exceptions.py
```diff
@@ -234,6 +234,79 @@ def __init__(self, *, file_id: UUID4, status_code: int = 400):
...
+                "box_id": str(box_id),
+                "max_size": max_size,
+                "current_size": current_size,
+            },
+        )
+
+
+class HttpBoxSizeLimitExceededError(HttpCustomExceptionBase):
+    """Thrown when adding a file would exceed the box's total size limit."""
+
```
Cito, line 274: maybe HttpBoxMaxSizeExceededError?

services/ucs/src/ucs/adapters/inbound/fastapi_/rest_models.py
```diff
@@ -27,19 +27,36 @@ class BoxCreationRequest(BaseModel):
...
 
 
 class BoxUpdateRequest(BaseModel):
     """Request body for updating a FileUploadBox."""
 
-    state: UploadBoxState = Field(default=..., description="Updated state")
+    state: UploadBoxState | None = Field(default=None, description="Updated state")
+    max_size: PositiveInt | None = Field(
+        default=None,
+        description="Updated maximum total bytes allowed across all file uploads in this box.",
```
Cito, line 43:
```suggestion
        description="Updated maximum total bytes allowed across all file uploads in this box (can only be enlarged).",
```

services/ucs/src/ucs/adapters/inbound/fastapi_/http_exceptions.py
Cito: of course, when the class name is changed, the id should also change here and in _routes, and the UploadController exceptions should also be adapted.

services/ucs/src/ucs/adapters/inbound/fastapi_/routes.py
```diff
@@ -193,39 +211,52 @@ async def update_box(
...
-    if (
-        work_order.box_id != box_id
-        or not required_work_type
-        or work_order.work_type != required_work_type
-    ):
+    if box_update.state is not None:
+        required_work_type = BOX_STATE_TO_WORK_TYPE.get(box_update.state)
+        if not required_work_type:
+            raise http_exceptions.HttpNotAuthorizedError()
+    else:
```
```suggestion
    else:  # the validator guarantees that max_size is set in this case
```

Cito (approved):
Looks good. Maybe consider simpler names for the errors.

services/ucs/src/ucs/adapters/inbound/fastapi_/http_exceptions.py
TheByronHimes: I went with HttpBoxMaxSizeTooLowError. It's the only scenario where the value passes validation and is too low.

services/ucs/src/ucs/adapters/inbound/fastapi_/rest_models.py
TheByronHimes: I think it's entirely possible that the number has to be adjusted down after something like a typo or miscommunication.

Cito: Right, that would be misleading. The max size itself can be decreased, just not beyond the current size. I wanted to point out this restriction. But it's not important, let's leave it like it is.

services/ucs/src/ucs/adapters/inbound/fastapi_/routes.py
```diff
@@ -122,15 +122,15 @@
             "Exceptions by ID:"
             + "\n- boxSizeLimitExceeded: Adding this file would exceed the box's size limit."
         ),
-        "model": http_exceptions.HttpBoxSizeLimitExceededError.get_body_model(),
+        "model": http_exceptions.HttpBoxMaxSizeExceededError.get_body_model(),
```
Cito, line 125: here the key in ERROR_RESPONSES still needs to be adapted: boxSizeLimitExceeded -> boxMaxSizeExceeded

Cito (changes requested):

Cito (approved):

PR #172, opened 2026-05-06
Stop UCS from deleting interrogated objects (GSI-2347)

TheByronHimes:
A bug in UCS caused it to perform inbox cleanup using the object ID and bucket ID from the `InterrogationSuccess` event, which meant it targeted the re-encrypted object in the _interrogation_ bucket.


dontseyit (commented):
Thanku

dontseyit (approved):

PR #173, opened 2026-05-07
UCS: Concurrent upload cap (GSI-2337)

TheByronHimes:
This adds the config option `max_concurrent_uploads_per_box` with a default of 5. Unlike in the epic spec, this cannot be disabled - I don't see an operational reason to do that.

The core gains a `TooManyOpenUploadsError`, which is translated at the API layer to `HttpTooManyOpenUploadsError`, status code 429.

There's a test that verifies the error translation and another one that verifies the cap functions as expected.

In the core, the in-progress file uploads are counted by looking for `state="init"`, which occurs at the same time as we're tallying up the current box size. (The core first checks if the concurrent upload cap is exceeded before checking if file size would exceeded remaining box space. This order is arbitrary since both items are available at the same time.)

### DIFF
--- a/services/ucs/src/ucs/core/controller.py
+++ b/services/ucs/src/ucs/core/controller.py
@@ -265,12 +265,25 @@ class UploadController(UploadControllerPort):
         extra["storage_alias"] = storage_alias
         extra["bucket_id"] = bucket_id
 
+        # Get both box size + in progress size and the number of in progress files
         current_size = box.size
+        in_progress_count = 0
         async for upload in self._file_upload_dao.find_all(
             mapping={"box_id": box.id, "state": "init"}
         ):
             current_size += upload.decrypted_size
+            in_progress_count += 1
 
+        # Ensure that another upload is allowed at the moment
+        max_concurrent = self._config.max_concurrent_uploads_per_box
+        if max_concurrent > 0 and in_progress_count >= max_concurrent:
+            error = self.TooManyOpenUploadsError(
+                box_id=box.id, max_concurrent=max_concurrent
+            )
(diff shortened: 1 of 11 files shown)


services/ucs/src/ucs/core/controller.py
```diff
@@ -265,12 +265,25 @@ async def initiate_file_upload(
...
+        in_progress_count = 0
         async for upload in self._file_upload_dao.find_all(
             mapping={"box_id": box.id, "state": "init"}
         ):
             current_size += upload.decrypted_size
+            in_progress_count += 1
 
+        # Ensure that another upload is allowed at the moment
+        max_concurrent = self._config.max_concurrent_uploads_per_box
+        if max_concurrent > 0 and in_progress_count >= max_concurrent:
```
Cito:
```suggestion
        if in_progress_count >= max_concurrent:
```

It has a PositiveInt validator so must be > 0 anyway.

services/ucs/tests_ucs/unit/test_controller.py
```diff
@@ -1216,6 +1216,54 @@ async def test_finished_uploads_count_toward_limit(rig: JointRig):
...
+
+    # Test that we can now upload a file again
+    _ = await rig.controller.initiate_file_upload(
+        box_id=box_id,
+        alias="new_file",
+        decrypted_size=1,
+        encrypted_size=1,
+        part_size=PART_SIZE,
+    )
+
```
Cito, line 1265: add a check that another upload again triggers the error

Cito (approved):
Looks good, too minor suggestions.

PR #175, opened 2026-05-11
Validate part size (GSI-2353)

TheByronHimes:
Enforces part size constraints: `5 MiB <= part_size <= 5 GiB` and the implied part count must be no greater than 10,000. If any of these constraints are not met, the API returns `400 Bad Request`.

### DIFF
--- a/services/ucs/src/ucs/core/controller.py
+++ b/services/ucs/src/ucs/core/controller.py
@@ -16,6 +16,7 @@
 """Implements the UploadController class to manage file uploads"""
 
 import logging
+from math import ceil
 from typing import Any
 from uuid import uuid4
 
@@ -29,6 +30,7 @@ from hexkit.utils import now_utc_ms_prec
 from pydantic import UUID4
 
 from ucs.config import Config
+from ucs.constants import MAX_PART_COUNT, MAX_PART_SIZE, MIN_PART_SIZE
 from ucs.core.models import FileUpload, FileUploadBasics, FileUploadBox
 from ucs.ports.inbound.controller import UploadControllerPort
 from ucs.ports.outbound.dao import (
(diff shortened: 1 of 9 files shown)


services/ucs/src/ucs/core/controller.py
```diff
@@ -291,6 +297,14 @@ async def initiate_file_upload(
             log.error(error, extra=extra)
             raise error
 
+        # Ensure part size is okay - verify size & implied part count
+        implied_part_count = ceil(encrypted_size / part_size)
+        if (
+            not (MIN_PART_SIZE <= part_size <= MAX_PART_SIZE)
+            or implied_part_count > MAX_PART_COUNT
```
Cito:
```suggestion
        if (
            not (MIN_PART_SIZE <= part_size <= MAX_PART_SIZE)
            or ceil(encrypted_size / part_size) > MAX_PART_COUNT
```

behaves better when part_size is 0

services/ucs/tests_ucs/unit/test_controller.py
```diff
@@ -1275,6 +1276,38 @@ async def test_concurrent_upload_cap(rig: JointRig):
         )
 
 
+@pytest.mark.parametrize(
+    "part_size, encrypted_size, expect_error",
+    [
+        (PART_SIZE, ENCRYPTED_SIZE, False),
+        (MIN_PART_SIZE - 1, ENCRYPTED_SIZE, True),
```
Cito, line 1284:
```suggestion
        (0, ENCRYPTED_SIZE, True),
        (MIN_PART_SIZE - 1, ENCRYPTED_SIZE, True),
```

Cito (changes requested):
Looks good, one minor thing.

services/ucs/src/ucs/core/controller.py
TheByronHimes: The validator would reject a part size of 0, but you're right that there's no reason not to account for that. I moved the min and max size constraints into the pydantic validation, too.

Cito (approved):

PR #176, opened 2026-05-11
UCS: Verify object size during mpu completion (GSI-2364)

TheByronHimes:
This adds a check during upload completion which verifies the actual object size reported by S3 against the expected object size reported by the submitter at creation time. Mismatches result in a 400 response to the user. Mirroring the error flow for checksum mismatches, the file is left in place for the user to delete. If the user neglects to delete the rejected file themselves, the UCS cleanup job can delete it later (not yet implemented).

### DIFF
--- a/services/ucs/src/ucs/core/controller.py
+++ b/services/ucs/src/ucs/core/controller.py
@@ -450,6 +450,42 @@ class UploadController(UploadControllerPort):
             log.error(error, extra=extra)
             raise error
 
+    async def _verify_object_size(self, *, file_upload: FileUpload) -> None:
+        """Verify that the S3 object size matches the declared encrypted_size.
+
+        If the sizes don't match, marks the FileUpload as failed and raises
+        UploadSizeMismatchError.
+        """
+        file_id = file_upload.id
+        object_id = file_upload.object_id
+        try:
+            actual_size = await self._s3_client.get_object_size(
+                file_upload=file_upload, object_id=object_id
+            )
+        except S3ClientPort.UnknownStorageAliasError as err:
+            raise self.UnknownStorageAliasError(
+                storage_alias=file_upload.storage_alias
+            ) from err
+
+        if actual_size != file_upload.encrypted_size:
+            file_upload.state = "failed"
+            file_upload.state_updated = now_utc_ms_prec()
+            file_upload.failure_reason = "Actual object size didn't match expected size"
+            await self._file_upload_dao.update(file_upload)
+            log.info("Marked FileUpload %s as 'failed' due to size mismatch.", file_id)
+            error = self.UploadSizeMismatchError(file_id=file_id)
+            log.error(
+                error,
+                extra={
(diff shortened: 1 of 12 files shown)


services/ucs/src/ucs/core/controller.py
```diff
@@ -450,6 +450,42 @@ async def _compare_checksums(
...
+        If the sizes don't match, marks the FileUpload as failed and raises
+        UploadSizeMismatchError.
+        """
+        file_id = file_upload.id
+        object_id = file_upload.object_id
+        try:
+            actual_size = await self._s3_client.get_object_size(
+                file_upload=file_upload, object_id=object_id
+            )
+        except S3ClientPort.UnknownStorageAliasError as err:
```
Cito, line 465: might also raise other errors - should we translate them?

services/ucs/src/ucs/core/controller.py
```diff
@@ -450,6 +450,42 @@ async def _compare_checksums(
...
+            raise self.UnknownStorageAliasError(
+                storage_alias=file_upload.storage_alias
+            ) from err
+
+        if actual_size != file_upload.encrypted_size:
+            file_upload.state = "failed"
+            file_upload.state_updated = now_utc_ms_prec()
+            file_upload.failure_reason = "Actual object size didn't match expected size"
+            await self._file_upload_dao.update(file_upload)
+            log.info("Marked FileUpload %s as 'failed' due to size mismatch.", file_id)
```
Cito: is the info needed if we already have the error?

services/ucs/tests_ucs/unit/test_controller.py
```diff
@@ -818,6 +818,56 @@ async def do_error(*args, **kwargs):
...
+        encrypted_size=ENCRYPTED_SIZE,
+        part_size=PART_SIZE,
+    )
+
+    file_upload = rig.file_upload_dao.latest
+    assert file_upload.state == "init"
+    state_updated = file_upload.state_updated
+
+    # Sleep so we can check for the timestamp difference
+    await sleep(0.1)
```
Cito: You should replace this (appears multiple times) with `sleep(1e-3)` corresponding to our ms resolution.

Cito (approved):
Some suggestions for improvement, but looks good already.

services/ucs/src/ucs/core/controller.py
TheByronHimes: Probably not - it's gone

services/ucs/src/ucs/core/controller.py
TheByronHimes: They should be handled, yes. I started to change that in this PR yesterday but realized it was going to inflate the LOC quite a bit because there are changes in several places and other errors that need to be handled too. I will fix that in a dedicated PR so it's not bound up with unrelated items

Cito (approved):

PR #177, opened 2026-05-12
FIS: Filter DHFS versions (GSI-2377)

TheByronHimes:
This adds a way to set the allowed DHFS versions through configuration using the `supported_dhfs_versions` parameter.

Bumps the FIS version to `11.1.0`. The monorepo version is not bumped because the currently set version (`12.0.0`) hasn't been released yet.

### DIFF
--- a/services/fis/src/fis/adapters/inbound/fastapi_/routes.py
+++ b/services/fis/src/fis/adapters/inbound/fastapi_/routes.py
@@ -36,7 +36,7 @@ def check_and_log_user_agent_header(
     *,
     request: Request,
     endpoint_description: str,
-    supported_dhfs_versions: str,
+    dhfs_version_constraint: str,
     storage_alias: str = "",
 ) -> None:
     """Log a structured message from the User-Agent header.
@@ -49,12 +49,18 @@ def check_and_log_user_agent_header(
     in a 426 Upgrade Required response.
     """
     raw_user_agent = request.headers.get("user-agent")
-    extra = {"user_agent": raw_user_agent, "storage_alias": storage_alias}
+    extra = {
+        "user_agent": raw_user_agent,
+        "storage_alias": storage_alias,
+        "dhfs_version_constraint": dhfs_version_constraint,
+    }
     alias_part = f", originating from {storage_alias}" if storage_alias else ""
     client_part = raw_user_agent
 
     accepted = True
     parts = (raw_user_agent or "").split("/")
+
+    # Note: this is a naive implementation and we might want/need to use regex later
     if (len(parts) == 2) and all(parts):
         extra["client_name"] = parts[0]
         extra["client_version"] = parts[1]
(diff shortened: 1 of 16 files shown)


services/fis/src/fis/adapters/inbound/fastapi_/routes.py
```diff
@@ -18,45 +18,80 @@
...
+        client_part = f"from {parts[0]}, version {version_part}"
+
+        # If the User Agent specifies DHFS, enforce the version requirement
+        if parts[0] == DHFS_USER_AGENT_PREFIX:
+            try:
+                accepted = Version(version_part) in SpecifierSet(
+                    supported_dhfs_versions
+                )
+            except InvalidVersion:
+                accepted = False
```
Cito, line 79:
```suggestion
                pass
```

Cito: Maybe add a comment that InvalidSpecifier is already covered by config validator.

services/fis/src/fis/adapters/inbound/fastapi_/routes.py
```diff
@@ -18,45 +18,80 @@
...
+        log.info(
+            "Received request %s for the %s endpoint%s.",
+            client_part,
+            endpoint_description,
+            alias_part,
+            extra=extra,
+        )
+    else:
+        log.warning(
+            "Rejected request from unsupported DHFS version %s.",
```
Cito: Better also add the supported_dhfs_versions in the warning so that you immedately see what was expected.

services/fis/src/fis/config.py
```diff
@@ -58,3 +59,25 @@ class Config(
             }
         ],
     )
+
+    supported_dhfs_versions: str = Field(
```
Cito: Consider making this singular (version) to make clear we expect a single string. Or maybe even clearer, something like `..._version_spec` or `..._version_range` or `..._version_constraint`.

services/fis/src/fis/constants.py
```diff
@@ -20,3 +20,4 @@
 TRACER = trace.get_tracer_provider().get_tracer(SERVICE_NAME)
 AUTH_CHECK_CLAIMS = ["iss", "aud", "iat", "exp", "sub"]
 GHGA = "GHGA"
+DHFS_USER_AGENT_PREFIX = "DataHubFileService"
```
Cito:
```suggestion
DHFS_USER_AGENT_PREFIX = "GHGA DataHubFileService"
```

Maybe we should make it more specific and help firewall ops understand what this is.

Cito (changes requested):

services/fis/src/fis/config.py
TheByronHimes: It is now `dhfs_version_constraint`

services/fis/src/fis/constants.py
TheByronHimes: Done. Will add this to the DHFS side too

Cito (approved):

PR #180, opened 2026-05-19
IFRS - fix v3 migration wrt bucket IDs

TheByronHimes:
The v3 migration did not account for the fact that old events lack `bucket_id`, and the test missed it because the mock event _did_ have that field. This patch pulls the bucket name from configuration for each storage alias and uses that in the migration to assign the correct value.


dontseyit (approved):
Looks good to me

PR #181, opened 2026-05-19
DCS: Add V3 Migration (GSI-2384)

TheByronHimes:
This adds the missing migration for DCS's persisted events and the core object. 

There were four event types, three of which were persisted, and only two of which were ever used. 
- FileDownloadServed ([old schema](https://github.com/ghga-de/ghga-event-schemas/blob/76f7464079e85c8168977c149a376275eeaddd58/src/ghga_event_schemas/pydantic_.py#L340), [new schema](https://github.com/ghga-de/ghga-event-schemas/blob/0e1ba03b3626ade0534ee6ac5ee274892fc7123a/src/ghga_event_schemas/pydantic_.py#L303))
  - `"file_id"` converted to uuid4 via same hashing mechanism as in IFRS's V3 migration
  - `"s3_endpoint_alias"` is renamed to `"storage_alias"`
- FileRegisteredForDownload ([old schema](https://github.com/ghga-de/ghga-event-schemas/blob/76f7464079e85c8168977c149a376275eeaddd58/src/ghga_event_schemas/pydantic_.py#L292), [new schema](https://github.com/ghga-de/ghga-event-schemas/blob/0e1ba03b3626ade0534ee6ac5ee274892fc7123a/src/ghga_event_schemas/pydantic_.py#L255))
  - `"upload_date"` is renamed to `"archive_date"`
  - Convert the `"file_id"` field's GHGA accession value to a uuid4 (actual UUID4, not a string representation of a uuid4) using same mechanism as in the IFRS v3 migration
  - `drs_uri` is removed
  - any other fields within the payload are left unchanged
For both persisted events, the `_id` field is updated so its accession portion to the right of the colon is replaced with the hash-derived UUID4. Same goes for the event key.

Then the DRS Objects collection is updated like so:
- the string accession (GHGA...) in `"_id"` is converted like above
- the field `"decryption_secret_id"` is renamed to just `"secret_id"`
- the field `"s3_endpoint_alias"` is renamed to `"storage_alias"`

### DIFF
--- a/services/dcs/src/dcs/migrations/definitions/v3.py
+++ b/services/dcs/src/dcs/migrations/definitions/v3.py
@@ -1,0 +1,128 @@
+# Copyright 2021 - 2025 Universität Tübingen, DKFZ, EMBL, and Universität zu Köln
+# for the German Human Genome-Phenome Archive (GHGA)
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Migration Definition for moving to DB version 3."""
+
+from contextlib import suppress
+from hashlib import sha256
+from uuid import UUID
+
+from hexkit.providers.mongodb.migrations import Document, MigrationDefinition
+from hexkit.providers.mongokafka.provider.persistent_pub import PersistentKafkaEvent
+
+from dcs.constants import DCS_PERSISTED_EVENTS_COLLECTION, DRS_OBJECTS_COLLECTION
+from dcs.core.models import AccessTimeDrsObject
+
+
+def derive_file_id_from_accession(accession: str) -> UUID:
+    """Use the first portion of the SHA256 hash of an accession to derive a UUID4"""
+    hash = sha256(accession.encode()).hexdigest()
+    uuid_str = f"{hash[0:8]}-{hash[8:12]}-4{hash[13:16]}-a{hash[17:20]}-{hash[20:32]}"
+    return UUID(uuid_str)
+
+
+async def update_drs_object(doc: Document) -> Document:
+    """Update a drs_objects document.
+
+    If _id is already a UUID, the doc has been migrated; skip it.
+    """
+    if isinstance(doc["_id"], UUID):
+        return doc
+
+    doc["_id"] = derive_file_id_from_accession(doc["_id"])
+
+    if "decryption_secret_id" in doc:
+        doc["secret_id"] = doc.pop("decryption_secret_id")
+
+    if "s3_endpoint_alias" in doc:
+        doc["storage_alias"] = doc.pop("s3_endpoint_alias")
+
+    return doc
+
+
+async def update_event(doc: Document) -> Document:
+    """Update a persisted event document.
+
+    If doc["key"] is already a valid UUID, the doc has been migrated; skip it.
+    Fields are updated field-by-field so this function handles any event type.
+    """
+    with suppress(ValueError):
+        _ = UUID(doc["key"])
+        return doc
+
+    payload = doc["payload"]
+
+    uuid4_file_id = derive_file_id_from_accession(payload["file_id"])
+    payload["file_id"] = uuid4_file_id
+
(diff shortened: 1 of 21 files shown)


services/dcs/src/dcs/migrations/definitions/v3.py
```diff
@@ -0,0 +1,128 @@
...
+
+from dcs.constants import DCS_PERSISTED_EVENTS_COLLECTION, DRS_OBJECTS_COLLECTION
+from dcs.core.models import AccessTimeDrsObject
+
+
+def derive_file_id_from_accession(accession: str) -> UUID:
+    """Use the first portion of the SHA256 hash of an accession to derive a UUID4"""
+    hash = sha256(accession.encode()).hexdigest()
+    uuid_str = f"{hash[0:8]}-{hash[8:12]}-4{hash[13:16]}-a{hash[17:20]}-{hash[20:32]}"
+    return UUID(uuid_str)
```
Cito, line 37: Maybe add a comment that IFRS uses the same algorithm so that the DBs stay consistent.

`hash` is a built-in function, better use some other name.

services/dcs/src/dcs/migrations/definitions/v3.py
```diff
@@ -0,0 +1,128 @@
...
+    """Update a drs_objects document.
+
+    If _id is already a UUID, the doc has been migrated; skip it.
+    """
+    if isinstance(doc["_id"], UUID):
+        return doc
+
+    doc["_id"] = derive_file_id_from_accession(doc["_id"])
+
+    if "decryption_secret_id" in doc:
```
Cito, line 50: Wouldn't it be better to fail here instead of silently ignoring missing fields?

services/dcs/src/dcs/migrations/definitions/v3.py
```diff
@@ -0,0 +1,128 @@
...
+
+async def update_event(doc: Document) -> Document:
+    """Update a persisted event document.
+
+    If doc["key"] is already a valid UUID, the doc has been migrated; skip it.
+    Fields are updated field-by-field so this function handles any event type.
+    """
+    with suppress(ValueError):
+        _ = UUID(doc["key"])
+        return doc
```
Cito, line 70: Is there a reason why the two functions use different code to check this?

services/dcs/src/dcs/migrations/definitions/v3.py
```diff
@@ -0,0 +1,128 @@
...
+
+    payload = doc["payload"]
+
+    uuid4_file_id = derive_file_id_from_accession(payload["file_id"])
+    payload["file_id"] = uuid4_file_id
+
+    doc["_id"] = doc["_id"].replace(doc["key"], str(uuid4_file_id))
```
Cito:
```suggestion
    doc["_id"] = f"{doc['_id'].rsplit(':', 1)[0]}:{uuid4_file_id}"
```

Feels a bit safer to me.

services/dcs/src/dcs/migrations/definitions/v3.py
```diff
@@ -0,0 +1,128 @@
...
+    doc["key"] = str(uuid4_file_id)
+    doc["published"] = False
+
+    if "target_object_id" in payload and isinstance(payload["target_object_id"], str):
```
Cito: Same question as above regarding the if's.

services/dcs/src/dcs/migrations/definitions/v3.py
```diff
@@ -0,0 +1,128 @@
...
+        payload["target_object_id"] = UUID(payload["target_object_id"])
+
+    if "s3_endpoint_alias" in payload:
+        payload["storage_alias"] = payload.pop("s3_endpoint_alias")
+
+    if "upload_date" in payload:
+        payload["archive_date"] = payload.pop("upload_date")
+
+    payload.pop("drs_uri", None)
```
Cito: Again, expect this to exist?

Cito (approved):
See some comments below.

services/dcs/src/dcs/migrations/definitions/v3.py
TheByronHimes: There is already document validation performed after the update, so if a field is actually missing then it will be caught after this function returns. This is different from the persisted events update, which doesn't benefit from the same validation because the payload is nested and the schema effectively unknown from the perspective of the migration manager.

services/dcs/src/dcs/migrations/definitions/v3.py
TheByronHimes: Yes, and I've added a comment explaining why

services/dcs/src/dcs/migrations/definitions/v3.py
TheByronHimes: Only one event type will have it. I don't know how any irregularity would occur, but to be extra safe I've moved it in with a check that only applies to that kind of event.

Cito (approved):

PR #182, opened 2026-05-20
DCS V3 Migration: Handle mixed data type for target_object_id

TheByronHimes:
For FileDownloadServed events, the `"target_object_id"` field was not converted to a UUID in the V2 migration, even though the schema itself was. So after the switch, new instances had that field with a UUID value while older instances had the string value. This never caused a production error because of the way pydantic handles UUIDs represented as strings. 

The DCS v3 migration encountered an error because it assumed that all old rows were strings that needed to be converted to UUIDs.

Bumps DCS version to `10.0.1` and the monorepo version to `13.0.1`

### DIFF
--- a/.pyproject_generation/pyproject_custom.toml
+++ b/.pyproject_generation/pyproject_custom.toml
@@ -1,6 +1,6 @@
 [project]
 name = "fsb"
-version = "13.0.0"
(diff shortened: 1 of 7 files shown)


dontseyit (approved):
thanke

PR #184, opened 2026-05-22
Add DHI image variant

TheByronHimes:
Adds the DHI image and updated workflows to publish it alongside the normal image.


.github/workflows/push_to_docker/action.yaml
mephenor, line 1: Why not have two different actions instead and use them conditionally?
Easier to have one with shared inputs and outputs?

mephenor (commented):
Why the major version bump for all services?
Didn't some of the services already have minor or patch bumps for the current release?

.github/workflows/push_to_docker/action.yaml
TheByronHimes: No real reason. As discussed offline, let's move refinements there to a future PR (maybe an immediate one if it doesn't work 🙂)

mephenor (approved):

PR #185, opened 2026-05-27
UCS: Stale Upload Cleanup Job (GSI-2338)

TheByronHimes:
This PR adds a new UCS command, `perform-cleanup [--run-forever]`, that performs cleanup duties on all configured `inbox` buckets. If the `--run-forever` flag is set, then the job will repeat after every `cleanup_interval_minutes`. The job aborts multipart uploads if either of the following two conditions are met:
- no activity has been performed since a configurable cutoff
- the upload id isn't connected to any in-progress `FileUpload`

To track upload activity, there's a new `UploadActivity` model & DAO. This just has the `file_id` and `last_activity`. It is created in the same core method that creates the corresponding `FileUpload`. The `last_activity` field, a datetime, is updated every time a URL is requested during upload. In order to return the response as soon as possible, the `last_activity` update operation is deferred until after the response is returned, handled by FastAPI's BackgroundTasks.

Errors are logged but otherwise allowed to crash the job.


services/ucs/src/ucs/main.py
```diff
@@ -66,3 +68,23 @@ async def consume_events(run_forever: bool = True):
...
+    async with prepare_core(config=config) as controller:
+        is_first_run = True
+        while run_forever or is_first_run:
+            is_first_run = False
+            log.info("Running stale upload cleanup job.")
+            await controller.cleanup_stale_uploads()
+            log.info("Stale upload cleanup job complete.")
+            if run_forever:
+                log.info(
+                    "Will run again in %i minutes.", config.cleanup_interval_minutes
```
mephenor:
```suggestion
                    "Will run cleanup again in %i minutes.", config.cleanup_interval_minutes
```
Else there is some context missing

services/ucs/src/ucs/core/controller.py
```diff
@@ -356,6 +372,21 @@ async def initiate_file_upload(
...
+            log.warning(
+                "Activity entry unexpectedly exists for new file %s. Overwriting.",
+                file_id,
+                extra={"box_id": box_id, "alias": alias, "file_id": file_id},
+            )
+        except ResourceNotFoundError:
+            pass
+        await self._upload_activity_dao.upsert(
+            UploadActivity(file_id=file_id, last_activity=now_utc_ms_prec())
+        )
```
mephenor: That's an extra roundtrip just for the logging here...
What are the downstream implications of the upload activity already existing?
Might be worth to reconsider the IO overhead vs. the value of the information here

services/ucs/src/ucs/core/controller.py
```diff
@@ -409,6 +440,28 @@ async def get_part_upload_url(self, *, file_id: UUID4, part_no: int) -> str:
                 bucket_id=file_upload.bucket_id, s3_upload_id=s3_upload_id
             ) from err
 
+        return presigned_url
```
mephenor: `presigned_url` is not used anywhere, so the previous direct return should be ok.

services/ucs/src/ucs/core/controller.py
```diff
@@ -409,6 +440,28 @@ async def get_part_upload_url(self, *, file_id: UUID4, part_no: int) -> str:
...
+                UploadActivity(file_id=file_id, last_activity=now_utc_ms_prec())
+            )
+        except ResourceNotFoundError:
+            log.warning(
+                "Activity entry missing for file %s during URL issuance. Creating it.",
+                file_id,
+            )
+            await self._upload_activity_dao.insert(
+                UploadActivity(file_id=file_id, last_activity=now_utc_ms_prec())
+            )
```
mephenor, line 443: Same consideration as above:
The extra roundtrip exists just for logging purposes.
What would be the tradeoff of just upserting unconditionally?

services/ucs/src/ucs/ports/outbound/storage.py
```diff
@@ -70,10 +70,21 @@ def __init__(self, *, file_id: UUID4, s3_upload_id: str, bucket_id: str):
...
+            *,
+            s3_upload_id: str,
+            object_id: str,
+            bucket_id: str,
+            file_id: UUID4 | None = None,
+        ):
+            if file_id is not None:
+                subject = f"file ID {file_id}"
+            else:
+                subject = f"object {object_id}"
```
mephenor: Small nitpick:
Adjust the logic to remove the `not` in the if branch, i.e.:

```suggestion
            if file_id is None:
                subject = f"object {object_id}"
            else:
                subject = f"file ID {file_id}"
```

services/ucs/src/ucs/core/controller.py
```diff
@@ -993,3 +1059,145 @@ async def process_file_deletion_requested(self, *, file_id: UUID4) -> None:
...
+        cutoff: UTCDatetime,
+    ) -> list[FileUpload]:
+        """Return uploads from ongoing_uploads whose last activity is older than cutoff.
+
+        Falls back to the upload's `initiated` timestamp when no activity entry exists.
+        """
+        stale_uploads: list[FileUpload] = []
+        for upload in ongoing_uploads:
+            try:
+                activity = await self._upload_activity_dao.get_by_id(upload.id)
```
mephenor: More of a tangential note:
Would it make sense to also make that batchable on the DAO?
This seems like something that would scale with the number of active uploads, but could be converted to one call covering all IDs.

Probably doesn't matter for now, as that's not a hot path.

services/ucs/src/ucs/core/controller.py
```diff
@@ -993,3 +1059,145 @@ async def process_file_deletion_requested(self, *, file_id: UUID4) -> None:
...
+        falls back to comparing the FileUpload's `initiated` timestamp.
+        """
+        cutoff = now_utc_ms_prec() - timedelta(
+            hours=self._config.multipart_upload_ttl_hours
+        )
+        for storage_alias in self._config.object_storages:
+            await self._cleanup_stale_uploads_for_alias(
+                storage_alias=storage_alias, cutoff=cutoff
+            )
+
```
mephenor, line 1063: Helper methods could be reordered to match the call order, but that's more of a personal preference.

mephenor (changes requested):

services/ucs/src/ucs/core/controller.py
TheByronHimes: It shouldn't happen, meaning that those calls would all be pointless except in maybe testing. I removed it.

services/ucs/src/ucs/core/controller.py
TheByronHimes: I've switched the approach so it gets all the relevant UploadActivity entries at once and filters afterward.

services/ucs/src/ucs/core/controller.py
TheByronHimes: I can do that in a follow-up PR.

mephenor: More of a nice to have, not strictly necessary

mephenor (approved):

PR #189, opened 2026-06-03
UCS: Handle edge case in File Upload init (GSI-2396)

TheByronHimes:
We recently had something happen in Staging where Kafka was misconfigured and it produce an error during the Outbox DAO's publishing action. This was during FileUpload init. So the `"init"`-state FileUpload was sitting around in the database, and the S3 upload had been opened, but the Connector only saw the error and therefore tried again. But these retries failed because there was an `"init"`-state FileUpload for the specified box & alias already. At the moment, there's no deletion mechanism in the Connector (to be fixed soon), so this effectively killed the user's ability to upload that file. 

This update adds error handling for this scenario to mark the upload as `"failed"` so that the Connector may at least retry in the future. Other options considered and abandoned were: A) Check the db and suppress the error/allow processing to continue if we see that the updates were written successfully to the DB (this then opens up a lot of questions about consistent logic and should we do it for, e.g. completion as well). B) Instead of marking the upload as `"failed"`, delete it entirely. (Not a bad option, could also do, but don't see an advantage other than the next attempt has fewer DB I/O.)

The reduce further failure points during error handling, the S3 uploads are left open (we'll let the cleanup job take care of it).

Some other error handling in the vicinity was improved as well.

No version bump because the previous UCS PR is still not released.

### DIFF
--- a/services/ucs/src/ucs/core/controller.py
+++ b/services/ucs/src/ucs/core/controller.py
@@ -174,8 +174,8 @@ class UploadController(UploadControllerPort):
             #  the cleanup job takes care of it.
             extra = {"box_id": file_upload.box_id, "file_alias": file_upload.alias}
             log.error(
-                "Got an error while trying to insert FileUpload %s, for which it must"
-                + " be marked 'failed'. Error: %s",
+                "Encountered an error while inserting FileUpload %s; it will be marked"
+                + " as 'failed'. Error: %s",
                 file_upload.id,
                 err,
                 extra=extra,
@@ -193,16 +193,16 @@ class UploadController(UploadControllerPort):
                 if isinstance(mark_failed_err, DaoError):
                     # Database errors should be raised, not suppressed
                     log.error(
-                        "An error has occurred during FileUpload init which requires"
-                        + " that the FileUpload be marked 'failed'. While making that"
-                        + " DB write, we received this additional error: %s",
+                        "While marking FileUpload %s as 'failed', encountered another"
+                        + " database error: %s",
+                        file_upload.id,
                         mark_failed_err,
                         extra=extra,
                     )
                     raise mark_failed_err
                 else:
                     log.warning(
-                        "While marking FileUpload %s as 'failed', got another error."
+                        "While marking FileUpload %s as 'failed', encountered another error."
                         + " If it's from Kafka, this is fine - just fix Kafka and run"
                         + " the publish-all job. If it's not related to Kafka at all,"
                         + " then this warrants further investigation. Error: %s",
--- a/services/ucs/src/ucs/core/controller.py
+++ b/services/ucs/src/ucs/core/controller.py
@@ -29,6 +29,7 @@ from ghga_event_schemas.pydantic_ import (
 )
 from ghga_service_commons.utils.utc_dates import UTCDatetime
 from hexkit.protocols.dao import (
+    DaoError,
     MultipleHitsFoundError,
     NoHitsFoundError,
     UniqueConstraintViolationError,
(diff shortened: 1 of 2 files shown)


services/ucs/src/ucs/core/controller.py
```diff
@@ -145,10 +163,30 @@ async def _insert_file_upload(
...
+            await self._file_upload_dao.update(file_upload)
+
+            # This is INFO level because it's normal procedure for cleanup.
+            log.info(
+                "Marked FileUpload %s as 'failed' due to error during initiation.",
+                file_upload.id,
+                extra=extra,
+            )
+            # Re-raise the exception
+            raise
```
mephenor, line 220: Is the `file_upload_dao` an outbox publisher here and the idea is to just update the state and don't care about the Kafka part?

Is the publishing deferred or does it happen directly after the update, i.e. will we get the actual error further up in the stack trace, masked by another KafkaError, while other errors will show up directly as the actual error from reraising?

If so, the behavior should be documented to avoid confusion.

mephenor: What's the state here, mpu initialized, but nothing else happened so far?

TheByronHimes: > Is the file_upload_dao an outbox publisher here and the idea is to just update the state and don't care about the Kafka part?

Exactly. When we resolve the issue, the events can be finally published using the `publish-events` command (which should run on a regular basis). 

> Is the publishing deferred or does it happen directly after the update[?]

Nominally, publishing occurs immediately after the DB write. In this case, the attempt fails and is later reconciled by the `publish-events` job (in theory - this needs to be set up if not already in place).

> What's the state here, mpu initialized, but nothing else happened so far?

State: mpu initialized but not utilized, db record is current, but corresponding kafka event has not been fired. No other user-driven upload activities have occurred.

TheByronHimes: > will we get the actual error further up in the stack trace, masked by another KafkaError, while other errors will show up directly as the actual error from reraising?

Should get proper chaining now as of [d6ca6b0](https://github.com/ghga-de/file-services-backend/pull/189/commits/d6ca6b07fd607d63e7d12318661ef5c74f21b793). 

If it's just Kafka, then we squash the second error which we expect to see after the `upsert`. You'll just get `FirstKafkaError` in the traceback, and the second one will be logged as a warning. However, if the issue is something else, you'll get the following:
```bash
OriginalError: Error text

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "controller.py", line 202, in <module>
    raise mark_failed_err
  File "controller.py", line 190, in <module>
    await self._file_upload_dao.upsert(file_upload)
ErrorFromUpsertAttempt: Error2 text
```

services/ucs/src/ucs/core/controller.py
```diff
@@ -111,44 +122,105 @@ async def _insert_file_upload(
...
+                raise error from None  # don't need Unique* error in the trace
+        except Exception as err:
+            # This branch handles all other errors that *don't* signify an existing file
+            # If, e.g. kafka raises an error, delete the FileUpload so user can retry.
+            # To avoid more potential failure points, let active S3 upload linger until
+            #  the cleanup job takes care of it.
+            extra = {"box_id": file_upload.box_id, "file_alias": file_upload.alias}
+            log.error(
+                "Got an error while trying to insert FileUpload %s, for which it must"
+                + " be marked 'failed'. Error: %s",
```
mephenor: Could bring the phrasing of this error message in line with the next one.

mephenor (approved):

PR #190, opened 2026-06-05
FIS: Delete file uploads in response to outbox events (GSI-2397)

TheByronHimes:
This PR is part of reconciliation with UCS. When a user creates a `FileUpload` for a given real file, then cancels it or it fails, then starts a _new_ `FileUpload` for that same real file (alias), UCS deletes the first one from the database in order to allow the second one to be created.

Before: FIS receives the `"deleted"`-type outbox event and marks its local copy of the `FileUpload` as `"cancelled"`. 

Now: FIS receives the `"deleted"`-type outbox event and _deletes_ its local copy of the `FileUpload`. It also deletes the associated `InterrogationReport`, if it exists.

Bumps FIS version to `11.3.0`.

### DIFF
--- a/.pyproject_generation/pyproject_custom.toml
+++ b/.pyproject_generation/pyproject_custom.toml
@@ -1,6 +1,6 @@
 [project]
 name = "fsb"
-version = "13.2.1"
+version = "13.3.0"
 description = "File Services Backend - monorepo housing file services"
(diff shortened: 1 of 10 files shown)


dontseyit (approved):
Nothing to object

PR #191, opened 2026-06-08
UCS: Fix Manual File Deletion

TheByronHimes:
This PR fixes a bug with UCS's DELETE File Upload endpoint. Before, UCS would try to delete the S3 object if the state was anything other than `init`. The problem is that once the FileUpload moves to `interrogated`, the bucket ID and object ID on the FileUpload refer to a bucket and object that UCS has no power over. 
The fix makes it so that UCS only attempts the S3 object deletion if the state is `inbox`. Similarly, it only attempts the abortion of the unfinished upload if the state is `init`. 

Version has not been bumped because the current version, 12.2.1, still hasn't been released.

### DIFF
--- a/services/ucs/tests_ucs/unit/test_controller.py
+++ b/services/ucs/tests_ucs/unit/test_controller.py
@@ -550,7 +550,49 @@ async def test_delete_file_upload_when_box_locked(rig: JointRig):
     assert len(file_upload_dao.resources) == 1
 
 
-async def test_delete_file_upload_when_upload_missing(rig: JointRig):
+@pytest.mark.parametrize(
+    "state",
+    ["interrogated", "awaiting_archival", "archived", "cancelled", "failed"],
+    ids=["interrogated", "awaiting_archival", "archived", "cancelled", "failed"],
+)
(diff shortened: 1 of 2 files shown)


services/ucs/tests_ucs/unit/test_controller.py
```diff
@@ -550,7 +550,49 @@ async def test_delete_file_upload_when_box_locked(rig: JointRig):
...
+
+    monkeypatch.setattr(rig.s3_client, "delete_inbox_file", spy_delete_inbox_file)
+    monkeypatch.setattr(
+        rig.s3_client, "abort_multipart_upload", spy_abort_multipart_upload
+    )
+
+    # Call remove_file_upload() and verify that the S3Client's delete method wasn't called
+    await rig.controller.remove_file_upload(box_id=box_id, file_id=file_upload.id)
+
+    assert not s3_calls, (
```
Cito: Maybe add init and inbox as state params, and check that in these cases, you have the corresponding s3_call. Just to make sure the test works as expected.

Cito (approved):
Looks good. One suggestion regarding tests.

PR #192, opened 2026-06-09
UCS: Return FileUploads Regardless of State (GSI-2406)

TheByronHimes:
Previously only returned FileUploads that had finished being uploaded to the inbox. Now it returns unfinished files too (and files that changed to cancelled or failed before uploading).


lkuchenb (approved):
Thanks!

PR #193, opened 2026-06-09
UCS: Unknown FileUpload Deletion Returns 404 (GSI-2407)

TheByronHimes:
Returns 404 from UCS when a client requests deletion for a FileUpload that doesn't exist.

Bumps the monorepo & UCS major versions to `14.0.0` and `13.0.0`, respectively.


lkuchenb (approved):
LGTM, thank you!

PR #195, opened 2026-06-10
UCS: Enhance box locking (GSI-2417)

TheByronHimes:
Returns a dedicated 409 error with detail when locking the box if there are unfinished uploads.  
This can be overridden by specifying `force: True` in the request, which will **abort** any unfinished uploads and lock the box.

Fixes a couple bugs related to cancelled/failed files too.

Bumps the versions of UCS and the monorepo to `14.0.0` and `15.0.0`, respectively.

### DIFF
--- a/services/ucs/src/ucs/core/controller.py
+++ b/services/ucs/src/ucs/core/controller.py
@@ -857,29 +857,53 @@ class UploadController(UploadControllerPort):
         await self._file_upload_box_dao.update(box)
         log.info("Updated max_size for box %s to %s.", box_id, max_size)
 
-    async def lock_file_upload_box(self, *, box_id: UUID4, version: int) -> None:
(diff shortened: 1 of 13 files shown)


services/ucs/src/ucs/core/controller.py
```diff
@@ -857,29 +864,56 @@ async def update_box_max_size(
...
+            if init_uploads:
+                log.info(
+                    "Aborted %d in-progress upload(s) in box %s for forced lock.",
+                    len(init_uploads),
+                    box_id,
+                )
+        else:
+            incomplete_files_cursor = self._file_upload_dao.find_all(
+                mapping={"box_id": box_id, "state": "init"}
+            )
```
Cito: could be moved before the "if"

Cito (approved):

PR #196, opened 2026-06-15
UCS: Delete boxes (GSI-2409)

TheByronHimes:
Adds an endpoint to delete FileUploadBoxes.

Also adds missing WOT support for file deletion requests originating from RS.

Bumps UCS version to `14.1.0` and monorepo version to `15.1.0`.


services/ucs/src/ucs/adapters/inbound/fastapi_/http_authorization.py
```diff
@@ -78,10 +80,21 @@ def __init__(
             config=self.wps_auth_config,
             context_class=models.CloseFileWorkOrder,
         )
+        # File deletion is requested by the WPS (user-driven cancellation via the
+        #  connector) and by the RS (manual deletion by a Data Steward or user),
+        #  each signing with its own key, so both providers are needed.
         self.delete_file_provider = JWTAuthContextProvider(
```
Cito: would call this delete_file_wps_provider then for clarity

Cito (approved):
LGTM

PR #197, opened 2026-06-22
UCS: Delete orphaned objects from the inbox bucket (GSI-2435)

TheByronHimes:
Adds on to the UCS's cleanup job to delete objects whose object ID doesn't correspond to any `init` or `inbox` state `FileUpload`. 

Does not bump the UCS version because there are unreleased changes in main that already contain the requisite version bump.

### DIFF
--- a/services/ucs/src/ucs/adapters/outbound/s3.py
+++ b/services/ucs/src/ucs/adapters/outbound/s3.py
@@ -489,9 +489,7 @@ class S3Client(S3ClientPort):
         bucket_id, object_storage = self._get_bucket_and_storage(storage_alias)
         extra: dict[str, Any] = {"storage_alias": storage_alias, "bucket_id": bucket_id}
         try:
-            object_ids: list[str] = await object_storage.list_all_object_ids(
-                bucket_id=bucket_id
-            )
+            object_ids = await object_storage.list_all_object_ids(bucket_id=bucket_id)
         except ObjectStorageProtocol.BucketNotFoundError as err:
             error = S3ClientPort.BucketNotFoundError(bucket_id=bucket_id)
             log.error(error, extra=extra)
--- a/services/ucs/src/ucs/adapters/outbound/s3.py
+++ b/services/ucs/src/ucs/adapters/outbound/s3.py
@@ -474,3 +474,30 @@ class S3Client(S3ClientPort):
             op_name="get_all_multipart_uploads", bucket_id=bucket_id, extra=extra
         ):
             return await object_storage.get_all_multipart_uploads(bucket_id=bucket_id)
+
+    async def cleanup_orphaned_objects(
+        self, *, storage_alias: str, known_object_ids: set[str]
+    ):
+        """Cleans out all orphaned object IDs in the inbox bucket.
+
+        Raises:
+            `UnknownStorageAliasError` if the storage alias is not known.
+        """
+        bucket_id, object_storage = self._get_bucket_and_storage(storage_alias)
+        extra = {"storage_alias": storage_alias, "bucket_id": bucket_id}
+        with handle_bucket_and_general_s3_errors(
+            op_name="list_all_object_ids", bucket_id=bucket_id, extra=extra
+        ):
+            object_ids = await object_storage.list_all_object_ids(bucket_id=bucket_id)
+            orphaned_object_ids = set(object_ids) - known_object_ids
+            for object_id in orphaned_object_ids:
+                await object_storage.delete_object(
+                    bucket_id=bucket_id, object_id=object_id
+                )
+                log.info(
+                    "Deleted object %s from bucket %s in storage alias %s.",
+                    object_id,
+                    bucket_id,
+                    storage_alias,
+                    extra={"object_id": object_id, **extra},
+                )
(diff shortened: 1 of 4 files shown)


services/ucs/src/ucs/adapters/outbound/s3.py
```diff
@@ -474,3 +474,74 @@ async def list_all_multipart_uploads(self, *, storage_alias: str) -> dict[str, s
...
+        extra: dict[str, Any] = {"storage_alias": storage_alias, "bucket_id": bucket_id}
+        try:
+            object_ids = await object_storage.list_all_object_ids(bucket_id=bucket_id)
+        except ObjectStorageProtocol.BucketNotFoundError as err:
+            error = S3ClientPort.BucketNotFoundError(bucket_id=bucket_id)
+            log.error(error, extra=extra)
+            raise error from err
+
+        orphaned_object_ids = set(object_ids) - known_object_ids
+        deleted_ids = []
```
mephenor, line 504: This could use an early return here when `orphaned_object_ids` is empty, with a corresponding log message

services/ucs/src/ucs/adapters/outbound/s3.py
```diff
@@ -474,3 +474,74 @@ async def list_all_multipart_uploads(self, *, storage_alias: str) -> dict[str, s
...
+                missing_ids.append(object_id)
+            except ObjectStorageProtocol.ObjectStorageProtocolError as err:
+                log.error(
+                    "Unable to clean up object %s from bucket %s in storage alias %s"
+                    + " because of the following error: %s",
+                    object_id,
+                    bucket_id,
+                    storage_alias,
+                    str(err),
+                    extra=extra,
```
mephenor: Something like `{**extra, "object_id": object_id}` could be used to get rid of the setup/teardown of the `object_id` information.

Would be an extra copy, but it feels less clunky.

services/ucs/src/ucs/adapters/outbound/s3.py
```diff
@@ -474,3 +474,74 @@ async def list_all_multipart_uploads(self, *, storage_alias: str) -> dict[str, s
...
+                )
+                problem_ids.append(object_id)
+            else:
+                deleted_ids.append(object_id)
+
+        _ = extra.pop("object_id", None)
+        extra["deleted_count"] = len(deleted_ids)
+        extra["deleted_object_ids"] = deleted_ids
```
mephenor: Would move this closer to the context it's used in, i.e. after next `if` block

services/ucs/src/ucs/ports/outbound/storage.py
```diff
@@ -228,3 +228,13 @@ async def list_all_multipart_uploads(self, *, storage_alias: str) -> dict[str, s
...
+
+    @abstractmethod
+    async def cleanup_orphaned_objects(
+        self, *, storage_alias: str, known_object_ids: set[str]
+    ):
+        """Cleans out all orphaned object IDs in the inbox bucket.
+
+        Raises:
+            `UnknownStorageAliasError` if the storage alias is not known.
+        """
```
mephenor: Docstring diverges from the actual implementation, i.e. should document the same error types.

services/ucs/src/ucs/core/controller.py
```diff
@@ -1427,15 +1427,25 @@ async def _cleanup_stale_uploads_for_alias(
...
     ) -> None:
-        """Run stale upload cleanup for a single storage alias."""
-        # Fetch all in-progress uploads for this storage alias
-        ongoing_uploads = [
+        """Run stale upload cleanup for a single storage alias.
+
+        This aborts ongoing uploads that are too old or not tied to any FileUpload.
+        Orphaned objects are also deleted.
+        """
+        # Fetch all FileUploads for this storage alias that are still 'init' or 'inbox'
```
mephenor, line 1435: This could use a quick `why` here, as the the `inbox` part of that information is only needed at the end to pass it through to the cleanup code.

services/ucs/src/ucs/adapters/outbound/s3.py
```diff
@@ -474,3 +474,74 @@ async def list_all_multipart_uploads(self, *, storage_alias: str) -> dict[str, s
...
+        When finished, log basic stats for deleted objects.
+
+        Raises:
+            `UnknownStorageAliasError` if the storage alias is not known.
+            `BucketNotFoundError` if the configured bucket does not exist in S3.
+        """
+        bucket_id, object_storage = self._get_bucket_and_storage(storage_alias)
+        extra: dict[str, Any] = {"storage_alias": storage_alias, "bucket_id": bucket_id}
+        try:
+            object_ids = await object_storage.list_all_object_ids(bucket_id=bucket_id)
```
mephenor: What's the return type here, a set of `str`s?

mephenor (changes requested):

services/ucs/src/ucs/adapters/outbound/s3.py
TheByronHimes: a list of strs

services/ucs/src/ucs/adapters/outbound/s3.py
```diff
@@ -489,18 +489,24 @@ async def cleanup_orphaned_objects(
         bucket_id, object_storage = self._get_bucket_and_storage(storage_alias)
         extra: dict[str, Any] = {"storage_alias": storage_alias, "bucket_id": bucket_id}
         try:
-            object_ids = await object_storage.list_all_object_ids(bucket_id=bucket_id)
+            object_ids: list[str] = await object_storage.list_all_object_ids(
```
mephenor: No need for the extra type annotation here, was just wondering about the input being stringified instead of using UUIDs

mephenor (approved):

PR #198, opened 2026-06-22
UCS: Idempotence fix report handling (GSI-2348)

TheByronHimes:
Fixes a latent bug. When UCS gets an InterrogationSuccess event, it has to update it's local copy of the `FileUpload` info in the DB and it has to delete the associated object from the database. When it does the DB update, it updates FileUpload's bucket_id and object_id to reflect the details of the data in the interrogation bucket. In the pre-patched version, UCS did this DB update _before_ deleting the S3 object from the inbox. We encountered a config issue that precipitated an error in the S3 deletion operation, and when we requeued the event we received a permission error because the information UCS used for the deletion operation now pointed to the interrogation bucket instead of the inbox. This patch reverses the order of operations so S3 deletion precedes the DB update.

Does not bump the UCS version because there are unreleased changes in main that already contain the requisite version bump.

### DIFF
--- a/services/ucs/tests_ucs/integration/test_controller.py
+++ b/services/ucs/tests_ucs/integration/test_controller.py
@@ -639,7 +639,7 @@ async def test_file_interrogation_report_happy(joint_fixture: JointFixture):
         s3_storage,
     )
 
-    # REGRESSION TEST: Publish/consume the InterrogationSuccess event to trigger S3 error
+    # Publish/consume the InterrogationSuccess event to trigger S3 error
     await joint_fixture.kafka.publish_event(
         payload=interrogation_success.model_dump(mode="json"),
         type_=config.interrogation_success_type,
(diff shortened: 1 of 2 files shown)


services/ucs/tests_ucs/integration/test_controller.py
```diff
@@ -626,7 +626,41 @@ async def test_file_interrogation_report_happy(joint_fixture: JointFixture):
...
+    await s3_storage.does_object_exist(bucket_id=inbox_bucket_id, object_id=object_id)
+    file_upload_check = file_upload_collection.find_one({"_id": file_id})
+    assert file_upload_check is not None
+    assert file_upload_check["state"] == "inbox"
+    assert file_upload_check["object_id"] == UUID(object_id)
+
+    # REGRESSION TEST: Undo the S3 patches. Regression test finished
+    s3_storage.delete_object = _real_delete_fn
+    controller._s3_client._get_bucket_and_storage = _real_get_storage_fn
+
```
mephenor, line 661: The `REGRESSION_TEST` prefix in the comments is weird.
Might make sense to keep that at the top, but drop it everywhere else.

TheByronHimes: I added it because in the wall of text I found it easier to keep track of which steps were specific to that sub-test if each comment had that prefix. I can remove it though

mephenor: Maybe change it to something like: REGRESSION_TEST_START and REGRESSION_TEST_STOP, so it's still clear what's part of it and drop the rest in between

mephenor (approved):

PR #200, opened 2026-06-23
UCS: Pagination (GSI-2408)

TheByronHimes:
Since boxes can potentially contain thousands of files, there needs to be a way to limit how many files UCS returns at once. To that end, this PR adds pagination to the GET endpoint which lists file uploads in a given box.

Bumps UCS version to `15.0.0` and the monorepo version to `16.0.0`.

### DIFF
--- a/services/ucs/src/ucs/adapters/inbound/fastapi_/http_exceptions.py
+++ b/services/ucs/src/ucs/adapters/inbound/fastapi_/http_exceptions.py
@@ -450,6 +450,20 @@ class HttpNotAuthorizedError(HttpCustomExceptionBase):
         )
 
 
+class HttpSkipOrLimitInvalidError(HttpCustomExceptionBase):
+    """Thrown when the skip or limit pagination parameters are invalid."""
+
+    exception_id = "skipOrLimitInvalid"
+
+    def __init__(self, *, status_code: int = 422):
+        """Construct message and initialize exception."""
+        super().__init__(
+            status_code=status_code,
+            description="The skip or limit pagination parameters are invalid.",
+            data={},
+        )
+
+
 class HttpInternalError(HttpCustomExceptionBase):
     """Thrown for otherwise unhandled exceptions"""
 
--- a/services/ucs/src/ucs/adapters/inbound/fastapi_/http_exceptions.py
+++ b/services/ucs/src/ucs/adapters/inbound/fastapi_/http_exceptions.py
@@ -450,16 +450,16 @@ class HttpNotAuthorizedError(HttpCustomExceptionBase):
         )
 
 
-class HttpSkipOrLimitInvalidError(HttpCustomExceptionBase):
+class HttpPaginationError(HttpCustomExceptionBase):
     """Thrown when the skip or limit pagination parameters are invalid."""
 
-    exception_id = "skipOrLimitInvalid"
+    exception_id = "paginationError"
 
     def __init__(self, *, status_code: int = 422):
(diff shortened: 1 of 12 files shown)


services/ucs/src/ucs/adapters/inbound/fastapi_/http_exceptions.py
```diff
@@ -450,6 +450,20 @@ def __init__(self, *, status_code: int = 403):
         )
 
 
+class HttpSkipOrLimitInvalidError(HttpCustomExceptionBase):
```
Cito: Maybe just HttpInvalidPaginationError?

services/ucs/src/ucs/adapters/inbound/fastapi_/routes.py
```diff
@@ -316,10 +322,11 @@ async def update_box(  # noqa: C901, PLR0912
     summary="Retrieve list of file IDs for box",
     operation_id="getBoxUploads",
     status_code=status.HTTP_200_OK,
-    response_model=list[FileUpload],
-    response_description="List of file IDs for completed uploads in the box",
+    response_model=rest_models.BoxUploadsPage,
+    response_description="Paginated list of file uploads for the box",
     responses={
         status.HTTP_404_NOT_FOUND: ERROR_RESPONSES["boxNotFound"],
+        status.HTTP_422_UNPROCESSABLE_CONTENT: ERROR_RESPONSES["skipOrLimitInvalid"],
```
Cito: I think the pydantic validation for skip and limit will be raised before the controller is even called, so we will always get the default validation error, not our own. we can keep it, but in openapi we should probably not advertize it.

services/ucs/src/ucs/adapters/inbound/fastapi_/routes.py
```diff
@@ -316,10 +322,11 @@ async def update_box(  # noqa: C901, PLR0912
     summary="Retrieve list of file IDs for box",
```
```suggestion
    summary="Retrieve all file IDs for the given box",
```

Since we're not returning a pure array anymore.

services/ucs/tests_ucs/unit/test_api.py
```diff
@@ -178,6 +180,87 @@ async def test_view_box_endpoint_auth(config: ConfigFixture, app_fixture: AppFix
...
+    """Test that the endpoint returns the correct paginated response shape and
+    forwards skip/limit query parameters to the controller correctly.
+    """
+    rs_jwk = config.rs_jwk
+    rest_client = app_fixture.rest_client
+    core_mock = app_fixture.core_mock
+    token_header = utils.view_file_box_token_header(jwk=rs_jwk, box_id=TEST_BOX_ID)
+    url = f"/boxes/{TEST_BOX_ID}/uploads"
+
+    # Create two file uploads with different aliases
```
Cito:
```suggestion
    # Build two mock file uploads with different aliases
```

Cito (approved):
Looks good. Just a few nitpicks.

services/ucs/src/ucs/adapters/inbound/fastapi_/routes.py
TheByronHimes: And it's also not the IDs but rather the whole file upload

services/ucs/src/ucs/adapters/inbound/fastapi_/http_exceptions.py
TheByronHimes: Changed to `HttpPaginationError`

PR #202, opened 2026-06-30
UCS: Add overwrite option for file uploads to replace active uploads (GSI-2460)

TheByronHimes:
Adds an `overwrite` parameter to UCS's `POST /boxes/{box_id}/uploads` endpoint. When used, UCS will attempt to cancel any existing upload in the box for the same alias specified in the request.

No version bump included because the unreleased changes in `main` already include a major version bump.

### DIFF
--- a/services/ucs/tests_ucs/unit/test_controller.py
+++ b/services/ucs/tests_ucs/unit/test_controller.py
@@ -1371,7 +1371,7 @@ async def test_overwrite_cancels_active_upload(
     "state",
     ["interrogated", "awaiting_archival", "archived"],
 )
-async def test_overwrite_blocked_for_terminal_states(
+async def test_overwrite_blocked_for_immutable_states(
     rig: JointRig,
     state: FileUploadState,
 ):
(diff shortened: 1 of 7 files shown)


services/ucs/tests_ucs/unit/test_controller.py
```diff
@@ -1303,6 +1303,160 @@ async def patched_insert(dto):
...
+
+    with pytest.raises(UploadControllerPort.FileUploadAlreadyExists):
+        await controller.initiate_file_upload(
+            box_id=box_id,
+            alias="test_file",
+            decrypted_size=DECRYPTED_SIZE,
+            encrypted_size=ENCRYPTED_SIZE,
+            part_size=PART_SIZE,
+            overwrite=True,
+        )
```
mephenor, line 1411: Need a bit more context here to see if I understand this correctly:
This is a pre-existing failure mode that would have previously been triggered by any state, but now it passes for `init` and `inbox`, as they call `remove_file_upload` first?

services/ucs/tests_ucs/unit/test_controller.py
```diff
@@ -1303,6 +1303,160 @@ async def patched_insert(dto):
...
+    assert (await file_upload_dao.get_by_id(file_upload.id)).state == state
+
+
+async def test_overwrite_false_still_blocks_active_upload(rig: JointRig):
+    """With overwrite=False (default), an active 'init' upload still raises
+    FileUploadAlreadyExists, confirming overwrite is opt-in.
+
+    The InMemDao doesn't enforce compound unique indexes, so we simulate the
+    MongoDB UniqueConstraintViolationError the same way existing idempotence tests do.
+    """
```
mephenor: Shouldn't this also cover `inbox` to be exhaustive?

TheByronHimes: Probably. I've added that

services/ucs/tests_ucs/unit/test_controller.py
TheByronHimes: The test (`test_overwrite_blocked_for_immutable_states`) makes sure that `overwrite=True` doesn't give the user superpowers enabling them to overwrite files that are now considered immutable.

mephenor (approved):

PR #208, opened 2026-07-08
UCS: Speedy box stats (GSI-2489)

TheByronHimes:
The FileUpload DAO was a really inefficient tool to recalculate current FileUploadBox size, so there is now an aggregator class that uses a Pymongo client directly. Error handling is minimal: the core class just catches whatever error, logs it, and re-raises it as `BoxStatsCalcError`. On the API side this is returned as a 500 without the extra logging that other random errors trigger.

### DIFF
--- a/services/ucs/src/ucs/adapters/outbound/box_stats.py
+++ b/services/ucs/src/ucs/adapters/outbound/box_stats.py
@@ -1,0 +1,59 @@
+# Copyright 2021 - 2026 Universität Tübingen, DKFZ, EMBL, and Universität zu Köln
+# for the German Human Genome-Phenome Archive (GHGA)
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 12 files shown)


services/ucs/src/ucs/adapters/outbound/box_stats.py
```diff
@@ -0,0 +1,59 @@
...
+                "$group": {
+                    "_id": None,
+                    "file_count": {"$sum": 1},
+                    "size": {"$sum": "$decrypted_size"},
+                }
+            },
+        ]
+        cursor = await self._collection.aggregate(pipeline)
+        async for group in cursor:
+            return group["file_count"], group["size"]
```
mephenor, line 58: Could use a comment that this will always return one group/document due to how `_id`:None is interpreted.

services/ucs/src/ucs/adapters/outbound/dao.py
```diff
@@ -81,7 +81,14 @@ async def get_file_upload_dao(self) -> DaoPublisher[FileUpload]:
...
                     fields={"box_id": 1, "alias": 1},
                     properties={"unique": True, "sparse": True},
-                )
+                ),
+                # Supports the stale upload cleanup job, which queries by
+                #  storage_alias and state
+                MongoDbIndex(fields={"storage_alias": 1, "state": 1}),
+                # Covering index for the box stats aggregation: matching on
+                #  box_id + state and summing decrypted_size can be served entirely
+                #  from this index without fetching documents
```
mephenor, line 90: A bit confused by the description:
A missing index wouldn't suddenly change the behavior of an aggregate, correct?
It's the aggregation that makes this run on the server without pulling the documents and evaluating locally.

mephenor (approved):

PR #209, opened 2026-07-09
UCS: Sort paginated uploads (GSI-2496)

TheByronHimes:
This is a follow-up PR for the recent pagination implementation. We add a `sort` parameter that expects a list of field names as a comma-delimited string. A validator checks each specified field name against the list of valid fields from the model. If not specified, then the default is to sort by `alias` ascending.

### DIFF
--- a/services/ucs/src/ucs/adapters/inbound/fastapi_/routes.py
+++ b/services/ucs/src/ucs/adapters/inbound/fastapi_/routes.py
@@ -325,7 +325,7 @@ async def update_box(  # noqa: C901, PLR0912
     },
 )
 @TRACER.start_as_current_span("routes.get_box_uploads")
-async def get_box_uploads(
+async def get_box_uploads(  # noqa: PLR0913
     box_id: UUID4,
     work_order: Annotated[
         rest_models.ViewFileBoxWorkOrder,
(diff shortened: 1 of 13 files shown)


services/ucs/src/ucs/adapters/inbound/fastapi_/routes.py
```diff
@@ -322,15 +325,24 @@ async def update_box(  # noqa: C901, PLR0912
...
     upload_controller: dummies.UploadControllerDummy,
     skip: Annotated[int, Query(ge=0)] = 0,
-    limit: Annotated[int, Query(ge=0, le=100)] = 50,
+    limit: Annotated[int, Query(ge=0, le=1000)] = 10,
+    sort: Annotated[
+        rest_models.SortString | None,
+        Query(
+            description="A comma-separated list of FileUpload field names defining"
+            + " the sort order, where field names prefixed with '-' indicate"
+            + " descending order (e.g. 'alias,-decrypted_size')."
```
```suggestion
            + " descending order (e.g. 'state,-decrypted_size')."
```

since the alias is unique the example would make no sense

services/ucs/src/ucs/adapters/inbound/fastapi_/routes.py
```diff
@@ -322,15 +325,24 @@ async def update_box(  # noqa: C901, PLR0912
...
+            + " Defaults to sorting by alias in ascending order."
```
Cito, line 343: Actually think we should always append `alias` to the sort list if it is not contained already, for consistency and to make the result stable. Otherwise you can get unexpected results with pagination when sorting just by state for instance.

services/ucs/src/ucs/adapters/outbound/box_stats.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+class MongoDbBoxStatsAggregator(BoxStatsAggregatorPort):
+    """Computes box stats with a server-side MongoDB aggregation.
+
+    This bypasses the hexkit DAO deliberately: the DAO would deserialize each matching
+    FileUpload document (including its large per-part checksum lists) into a Pydantic
+    model just to read `decrypted_size`. The aggregation instead sums the values inside
+    MongoDB and returns only the two resulting numbers.
+    """
```
Cito, line 32: One problem with this approach is that it is not really tested now - we only test with the InMemBoxStatsAggregator. I think it makes sense to include a real Mongo based integration test for the aggregation.

Cito (changes requested):
Looks good. Two suggestions.

services/ucs/src/ucs/adapters/outbound/box_stats.py
TheByronHimes: Thanks - I'll add this to the correct PR #208

PR #210, opened 2026-07-10
UCS: Sort Paginated FileUploads (GSI-2496)

TheByronHimes:
This is a follow-up PR for the recent pagination implementation. We add a `sort` parameter that expects a list of field names as a comma-delimited string. A validator checks each specified field name against the list of valid fields from the model. If not specified, then the default is to sort by `alias` ascending.


Cito (approved):
The only small issue is that this allows duplicate or contradictory sort specs (like alias,-alias), and sorting by internal fields that do not have an index. But I wouldn't over-engineer this as this  is more or less an internal API only. Let's keep it like that.

PR #211, opened 2026-07-13
UCS: Optional checksums (GSI-2492)

TheByronHimes:
This adds a `with_checksums` parameter to the `GET /boxes/{box_id}/uploads` endpoint. By default it is set to `False`, which is a breaking change. Since UCS version `15.0.0` is already on deck and unreleased, this is fine. The point of this change is to reduce the amount of data transferred to the frontend (the checksum lists can constitute a non-trivial amount of data, especially at higher part counts and file counts). This does _not_, however, prevent the data from being pulled back from MongoDB because of the limitations imposed by our DAO solution (mostly just a technical footnote).

There is a corresponding PR for the RS [here](https://github.com/ghga-de/ghga-registry-service/pull/22).

### DIFF
--- a/services/ucs/src/ucs/core/controller.py
+++ b/services/ucs/src/ucs/core/controller.py
@@ -1161,6 +1161,7 @@ class UploadController(UploadControllerPort):
         skip: int = 0,
         limit: int | None = None,
         sort: list[str] | None = None,
+        with_checksums: bool = False,
     ) -> tuple[list[FileUpload], int]:
         """Return a page of FileUploads for a FileUploadBox.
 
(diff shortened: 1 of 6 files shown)


services/ucs/src/ucs/core/controller.py
```diff
@@ -1197,7 +1201,16 @@ async def get_box_file_info(
         except ValueError as err:
             raise self.PaginationError() from err
 
-        file_uploads = [x async for x in find_result]
+        if with_checksums:
+            # TODO: Use `.to_list()` once newer hexkit is pulled in
+            file_uploads = [x async for x in find_result]
+        else:
+            file_uploads = [
```
Cito, line 1208: This is good since it does not materialize the full list in memory. But it still loads everything from the db over the network. We should consider adding a feature to hexkit that allows excluding certain fields (replacing them with a default value), which would use mongodb projection under the hood.

Cito (approved):
Looks solid to me.

services/ucs/src/ucs/core/controller.py
TheByronHimes: We should talk about it at some point. Every change like this mutates the DAO protocol into a more of a wrapper around the Pymongo client.

Cito: Yes, the danger exists. On the other hand, simple features such as exclusion of filelds are supported by all databases, and it would not be difficult to write providers for them. Would be good to add support for one very different database (maybe SQL) to force us to add only portable features.

